PR #1, opened 2020-11-09
Unwrap all i18n calls

CGastrell:
This PR just unwraps all our i18n calls so we use the `@wordpress/i18n` directly.

Once this is merged, next releases should be able to export our translatable strings to the .pot file and get into the wpcom translations flow/cron (ref: p1604357927213600-slack-C02ER2E0Z)

## Test instructions

Checkout and run:

  - `npm run lint:all`
  - `make release`

Verify the strings from JS files are extracted to the .pot file


CGastrell:
I should remove the now unused `lib/i18n` file

client/blocks/applause/sidebar.js
```diff
@@ -16,13 +16,12 @@ import {
...
 import { includes } from 'lodash';
-
-import { PollStatus } from './constants';
+import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
  */
-import { __ } from 'lib/i18n';
+import { PollStatus } from './constants';
```
ice9js, line 24: The world is a better place now!

client/blocks/poll/sidebar.js
```diff
@@ -311,114 +344,117 @@ const SideBar = ( {
...
-							label: __( 'Default theme font' ),
+							label: __(
+								'Default theme font',
+								'crowdsignal-forms'
+							),
 							value: FontFamilyType.THEME_DEFAULT,
 						},
 						{
-							label: __( 'Alegreya Sans' ),
+							label: __( 'Alegreya Sans', 'crowdsignal-forms' ),
```
ice9js, line 357: Why are we translating font names 👽  ? Let's remove this while we're at it.

ice9js (changes requested):
Nice! I don't think we need to be translating font names though 😄

ice9js (approved):
Nice! Thanks for cleaning up the old lib as well ;)

PR #3, opened 2020-11-11
Use default theme font for Applause count

CGastrell:
This PR changes the Applause count tag so it inherits the editor styles. It also compensates the margin assigned to `p` tags.

Tested well on:

  - Twenty Twenty Theme
  - Twenty Nineteen Theme
  - Twenty Seventeen Theme
  - Neve Theme
  - Seedlet Theme
  - Astra Theme

## Test instructions

Checkout and run `make client`. Add/edit a post with an Applause block. Verify the clap count text shows the default theme font both in the editor and in the public frontend

### DIFF
--- a/client/components/applause/index.js
+++ b/client/components/applause/index.js
@@ -95,9 +95,9 @@ const Applause = ( props ) => {
 				tabIndex={ 0 }
 			>
 				<ApplauseAnimation active={ animationActiveState } />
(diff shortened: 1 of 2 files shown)


digitalwaveride (approved):
As advertised!! :shipit:  
Thank you, Sir!

PR #4, opened 2020-11-12
Fix applause branding styles

CGastrell:
This PR adjusts the brand link shown on free users so that it:

  - compensates width of the link to better align with the block
  - removes shadow and border as those strategies are used to *underline* the link on some themes

## Test instructions

Checkou and `make client`. Visit a post with an applause block created from a free Crowdsignal account. Once clapped, the block should show the brand link below.
Verify the link looks centered.
Verify the link is not underlined


digitalwaveride (approved):
Sweet! 🚢 
Thanks Christian!

PR #5, opened 2020-11-17
Release v1.3.3

CGastrell:
This PR is a version bump including latest changes as well as some little fixes on the release flow:

  - invert order between `pot` and `client` Makefile tasks: trying to run the `make-pot` command after the JS files have been compiled leads to an overflow error
  - ignore directories on the `make-pot` command: as we actually don't need the command to dig into those directories, we might as well ignore them

## Test instructions

Verify occurrences of version have been bumped. Test locally to verify build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ client:
 	npm run build:applause
 
 # Package for release
-release: clean-release install-node client pot
+release: clean-release install-node pot client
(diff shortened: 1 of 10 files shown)


digitalwaveride (approved):
Things worked for me! Thanks Christian!

PR #6, opened 2020-12-02
Fix applause count style

CGastrell:
Applause count style is being overridden on P2s:
image

This PR adds a `margin: unset important!` to the count wrapper.
image

## Test instructions
Checkout and `make client`
See that the applause blocks look ok on local docker and on P2s


digitalwaveride (approved):
Thanks Christian!!! 🙏

PR #11, opened 2020-12-14
Show branding on editor

CGastrell:
This PR displays branding links while editing the poll/vote/applause blocks in order to show a consistent preview.

Brand link should now show exactly how it will be seen on the public post. A tooltip has been added on mouseover to invite the user to check the upgrade options:
image

## Test instructions:

NOTE: this PR depends on D54527-code which hasn't been deployed yet.

Checkout and run `make client`. With a free account, create/edit a post with a Crowdsignal Form's block. Shortly after inserting the block, the brand link should be viewable.


digitalwaveride:
Nice Christian! Great to see this.

A few finetuning nitty-gritty things:

**Hide Label Styling:**

```
.crowdsignal-forms-poll__footer-branding .crowdsignal-forms__branding-promote {
  background-color: #A4A4A4
  padding-top: 4px;
  padding-bottom: 4px;
  font-size: 10px;
}
```
After:

image


**Increasing spacing above the branding**

div.crowdsignal-forms-poll__footer-branding {
  margin-top: 16px;
}

After:

image


**Getting Label and Branding Text horizontally centered:**

image

Did not find on first sight how to tackle this... maybe an idea how to center this a bit more?


**Tooltip**

Also can't see the arrow, unfortunately.

image

Also there seems to be quite a delay till the tooltip shows, we might want to speed this up a bit in that case.

An the tooltip needs a bit more padding on all sides. Would suggest 12px.


**Applause Block & Voting Block**

The CSS is broken for me there:
image

image

BUT, we will not show the branding and hide labels there, as discussed, as they are too noisy, and will likely hurt adoption too much.

For these blocks we will have just the upgrade promotion in the side-bar.

digitalwaveride:
Christian, thanks so much for the update! Here a few things I noticed so far:

## Poll Block

That's what I am seeing currently:

image


### "Hide Label" Styling

Label styling is still off for me. -> that's what I get:

image

That's what we are looking for ... see above:

>   background-color: #A4A4A4
>   padding-top: 4px;
>   padding-bottom: 4px;
>   font-size: 10px;
>   color: #fff

### Link on "Hide Label"

Currently we have only a link in the tool tip, the label itself should have a link as well, the same link destination as the "upgrade" link in the tooltip.

### Positioning of the Hide Label

As mentioned above, any chance to center the label compared to the text?

image


### Signals Warning Font Styling

image

As mentioned on Slack, these notifications are not part of the theme, but kind of a "WP notification". We should use the same font styling as in "placeholders"/other notifications ... for example in the "Repeated visitor" block ... or in the Gutenberg sidebar.

image


image


### Sidebar Styling

image

The line-break with the "link icon" is a bit unfortunate. Any chance to tune that a bit? We certainly can skip the "!" after "Learn more", and decrease the margin on the left, but likely this won't be enough.
In case, let's get rid of the "Learn more", and instead put a link on "reached the signals limit".

## Applause & Voting Block

Everything is great there, besides the same sidebar line-break issue as mentioned above.

digitalwaveride:
Awesome Christian! We are very close, one little thing:
The font-size and font-height in the notification banner is still a bit off:

image

We should use the default styling for all these placeholders, notifications (see above the Amazon block as a random example)

```

font-size: 13px;
line-height: 23.4px
```

digitalwaveride:
🚀🚀🚀

digitalwaveride (approved):

PR #12, opened 2020-12-21
Remove redirect URL feature

### LINKED ISSUE #14: Redirect after voting might lead to phishing URLs
Poll blocks allow to redirect to a different URL after voting. This represents a security concern since a user could set there the URL of a phishing/scam site.

CGastrell:
This PR removes the "redirect to URL after voting" feature until further notice.

Existent polls with this option are now being forced into *thank you* message. The option is no longer available on the sidebar.

## Test instructions

Checkout and `make client`

Visit a post with some poll block with redirect option set. After voting, you should see the default "Thank you" message.

Create a post and add a poll block to it. You should not see the "Redirect to URL" option on the sidebar.

Closes #14

### DIFF
--- a/client/blocks/poll/sidebar.js
+++ b/client/blocks/poll/sidebar.js
@@ -146,6 +146,11 @@ const SideBar = ( {
 
 	const answerStyle = getAnswerStyle( attributes, className );
 
+	// Force redirect into simple THANK_YOU until some workaround to prevent phishing
+	if ( ConfirmMessageType.REDIRECT === attributes.confirmMessageType ) {
+		attributes.confirmMessageType = ConfirmMessageType.THANK_YOU;
+	}
+
 	return (
 		<InspectorControls>
 			<PanelBody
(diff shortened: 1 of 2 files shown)


client/blocks/poll/sidebar.js
```diff
@@ -146,6 +146,11 @@ const SideBar = ( {
 
 	const answerStyle = getAnswerStyle( attributes, className );
 
+	// Force redirect into simple THANK_YOU until some workaround to prevent phishing
```
ice9js, line 149: Can we maybe open a GitHub issue with a slightly better explanation for this and just link to it here instead?
Will probably make things easier for when we actually get around to this.

ice9js (approved):
Works as advertised. Thanks!

Before you merge, I'd recommend putting a short summary of this into a new issue and then just link to it from the source instead of all the `Suspended until some workaround to prevent phishing` comments.

client/blocks/poll/sidebar.js
CGastrell: I've created #12 with a succinct issue description, getting the link on the commented code now. We can further edit the issue if we feel like explaining more of it.

PR #13, opened 2020-12-21
Add popup & dialog logic for the NPS block

ice9js:
This patch adds the necessary logic for the block/dialog to be rendered on the frontend.

I've defined a new `viewThreshold` attribute which we'll be using to trigger the dialog at the right time. The setting might not save correctly right now due to there not being a `surveyId` (which will be addressed in another PR that integrates it with the Crowdsignal API), but the default value should be good enough for testing.

Another thing worth noting is, I've updated the server side `Crowdsignal_Forms_Nps_Block::should_hide_block()` to also hide the block unless it's a single page as it could become troublesome to manage if these started appearing on feeds.

The dialog is still a placeholder here and can be closed by clicking on the backdrop.

## Testing

- Add an 'NPS block' to your post and verify a dialog appears after visiting the page exactly three times.
- When viewing the post on a feed, the 'block' should be removed from the post content.

### DIFF
--- a/assets/stylesheets/nps.scss
+++ b/assets/stylesheets/nps.scss
@@ -1,3 +1,7 @@
 // Shared
 @import "./shared/animations.scss";
 @import "./shared/variables.scss";
+
+// Components
+@import "components/dialog-wrapper/style.scss";
+@import "components/nps/style.scss";
--- a/client/blocks/nps/attributes.js
+++ b/client/blocks/nps/attributes.js
@@ -13,4 +13,8 @@ export default {
 		type: 'string',
 		default: null,
 	},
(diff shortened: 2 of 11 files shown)


CGastrell (approved):
Works nicely! :) 

WRT not showing on feeds, I think that's a good idea. Where does that `is_single()` come from? Is it a "global" function on WP?

ice9js:
>WRT not showing on feeds, I think that's a good idea. Where does that is_single() come from? Is it a "global" function on WP?

Yeah, I found it [here](https://developer.wordpress.org/reference/functions/is_single/) - happy to consider alternatives if you know any :D

PR #15, opened 2020-12-28
Integrate the NPS block with the Crowdsignal API

ice9js:
This patch adds the capability to save changes made inside an NPS block to Crowdsignal.com.

For the time being, it's just through a simple (and ugly) save button at the top of the block.  
This is only a temporary solution while we continue discussions around synchronizing in pabtAt-1a5-p2. The button is a bare minimum and as such all of its code will be required by any of the proposed solutions which keeps our options open but allows us to move forward.

# Testing

- Create a new NPS block or open an existing one. You'll see a difficult to miss 'save' button.
- Click the save button, the block should now appear in the dashboard as a survey and a `surveyId` attribute should be set on the block.
- Change the block content, click the button again, Crowdsignal.com dashboard should be updated accordingly.

### DIFF
--- a/client/blocks/nps/attributes.js
+++ b/client/blocks/nps/attributes.js
@@ -5,13 +5,25 @@
  */
 
 export default {
+	feedbackQuestion: {
+		type: 'string',
+		default: '',
+	},
 	hideBranding: {
 		type: 'boolean',
 		default: false,
 	},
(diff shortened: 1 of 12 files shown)


CGastrell (approved):
LGTM!

PR #17, opened 2021-01-07
Update the Makefile to include the NPS block build

ice9js:
Small fix to include `npm run build:nps` in `make client`.

# Testing

Run `make client` and verify it builds `build/npm.js`.


CGastrell (approved):
LGTM!

PR #18, opened 2021-01-07
Refactor withClientId

ice9js:
I refactored this component as part of my initial effort to persist NPS block surveys.  
Although it seems like we'll be going in a different direction and it's no longer required for that, I feel like it's still worth including it.

I updated the signature so it always expects an array instead of an array or a string (varargs would be another solution) and moved it up a level so it can now be used correctly with `compose` or `flowRight` - the HoC signature should only ever accept one argument which is the component.

# Testing

This is a core change but it's quite easy to test.

- Create a post with some new poll, applause and vote blocks.
- Verify the blocks save correctly to the dashboard
- Publish the post and add some responses
- Verify the responses are recorded correctly as well

### DIFF
--- a/client/blocks/applause/edit.js
+++ b/client/blocks/applause/edit.js
@@ -46,6 +46,7 @@ const EditApplauseBlock = ( props ) => {
 	);
 };
 
-export default compose( [ withPollBase ] )(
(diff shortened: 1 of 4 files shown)


CGastrell (approved):
Looks good, works good.

PR #19, opened 2021-01-12
Implement public block views and voting interactions

ice9js:
This patch adds the initial public block views and voting interactions and updates the editor to allow for editing these elements as well.

Note that it's just the frontend. Actual vote recording will be implemented in a future patch.  
You should be able to edit all the strings on the public block except the rating buttons which are fixed.

The block will always take up 100% of the content width inside the editor and the dialog will match the theme's content width when displayed on a published post/page.

When you select a rating, the number should get highlighted briefly while the request is being handled (currently just a delay) and the feedback question should be shown soon after.  
The dialog should close itself after submitting the feedback question.

Screen Shot 2021-01-12 at 9 41 33 PM
Screen Shot 2021-01-12 at 9 41 47 PM

Screen Shot 2021-01-12 at 10 07 49 PM

# Testing

Run `make client` to update the assets.

Add a new NPS block and fill in the different fields using some example values. The editable fields include: rating and feedback question, rating scale labels, feedback submit button.

For the purpose of testing this PR, it's not necessary to save the block to the dashboard.

Publish/preview the post and visit the page three times (or what you set using the view threshold setting), the dialog should be displayed.  
Validate that the strings match the editor values and the interactions work correctly.

### DIFF
--- a/assets/stylesheets/editor.scss
+++ b/assets/stylesheets/editor.scss
@@ -19,3 +19,7 @@
 // Applause
 @import "blocks/applause/edit.scss";
 @import "components/applause/style.scss";
+
+// NPS
+@import "blocks/nps/edit.scss";
+@import "components/nps/style.scss";
--- a/client/blocks/nps/attributes.js
+++ b/client/blocks/nps/attributes.js
@@ -1,4 +1,9 @@
 /**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
  * Note: Any changes made to the attributes definition need to be duplicated in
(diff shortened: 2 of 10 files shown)


digitalwaveride:
As mentioned on Slack, no design specific review here.

> Publish/preview the post and visit the page three times (or what you set using the view threshold setting), the dialog should be displayed.
Validate that the strings match the editor values and the interactions work correctly.

Worked all for me as advertised!

> The dialog should close itself after submitting the feedback question.

That was not the case for me! Clicking submit on the feedback question did nothing for me. Modal stayed open, till I clicked the "x". (But that's maybe related to not yet saving anything?)

digitalwaveride (approved):

PR #20, opened 2021-01-14
Add support for submitting NPS responses

ice9js:
This patch further integrates the block with Crowdsignal by allowing to record NPS responses from the block.
A small side-effect of this PR is the block will now display correctly on every refresh when previewing a post or a page.

This time the new WP-API is a simple, almost 1:1 proxy for Crowdsignal's NPS response endpoint. The differences are the addition of a `checksum` field in the response, as well as the addition of a `nonce` field to the request.

Because the response is submitted in two steps and we're potentially allowing anyone to submit the requests to the WP API, it's important to protect ourselves from a case where it'd be possible for someone to alter someone else's response just by knowing the response ID.  
The solution for this is to use the response ID and the current respondent's nonce to create a checksum as soon as the rating is recorded that is to be submitted in all subsequent request.

# Testing

- Create an NPS block, fill in the important values, press the save button and publish the post.
- Verify the block uses the correct `viewThreshold` value when viewing the public page.
- Verify the block is displayed on every view when previewing the post.
- Try submitting a rating and closing the dialog. You should get a response for your survey in the Crowdsignal dashboard. The feedback should be empty.
- Try submitting a rating and feedback. The dialog should close after the feedback has been submitted. You should get a response with both the rating and the feedback inside the Crowdsignal dashboard.

Bonus: 
- Try manually submitting a feedback request using the responseID and checksum from another session. This should not work. Neither if you try to submit it using a valid nonce from a different session.

### DIFF
--- a/client/blocks/nps/attributes.js
+++ b/client/blocks/nps/attributes.js
@@ -35,7 +35,7 @@ export default {
 		default: __( 'Submit', 'crowdsignal-forms' ),
 	},
 	surveyId: {
-		type: 'string',
+		type: 'number',
 		default: null,
 	},
 	title: {
--- a/client/components/nps/feedback.js
+++ b/client/components/nps/feedback.js
@@ -1,16 +1,53 @@
 /**
  * External dependencies
  */
-import React from 'react';
+import React, { useState } from 'react';
 
-const NpsFeedback = ( { attributes } ) => (
-	<div className="crowdsignal-forms-nps__feedback">
-		<textarea className="crowdsignal-forms-nps__feedback-text" rows={ 6 } />
+/**
+ * Internal dependencies
+ */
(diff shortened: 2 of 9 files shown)


digitalwaveride:
Gave it another on run on Firefox... everything worked now regarding the saving behaviour as advertised! 🚀 

Two things I noticed, outside of the test-plan:

-> Voting a "0" is not counted! - I guess that is a known issue ... on the API side?

-> When showing the NPS block in preview mode, the block shows immediately 🤟  ... but the redirect after submitting the feedback question, does not forward me to the “post”, but it redirects to the “homepage” of the site. (that's not the case, when testing on a published post)

digitalwaveride (approved):

PR #21, opened 2021-01-19
Add NPS-specific toolbar options

ice9js:
This patch implements toolbar options specific to the NPS block. That is:

- Toggling the view between rating and feedback views.
- Setting the view threshold.

Note the icons are still only temporary.

Screen Shot 2021-01-18 at 10 29 27 PM

# Testing

- Add an NPS block, you should only see the rating view of the block. You can toggle between the rating and the feedback view using the 'Rating/Feedback' toggle in the toolbar.

- Click on the eye icon in the toolbar, you should see a popover with an input for the view threshold. Edit the number and verify your new value is saved and works correctly when the post is published.

### DIFF
--- a/client/blocks/nps/constants.js
+++ b/client/blocks/nps/constants.js
@@ -1,0 +1,4 @@
+export const views = {
+	RATING: 'rating',
+	FEEDBACK: 'feedback',
+};
--- a/client/blocks/nps/edit.js
+++ b/client/blocks/nps/edit.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import React from 'react';
+import React, { useState } from 'react';
 import { pick, times } from 'lodash';
 
 /**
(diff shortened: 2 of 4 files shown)


digitalwaveride (approved):
Based on the test plan, all worked as advertised!

Two things I noticed:
- In "Preview" mode, the NPS block did not show immediately, but needed to pass the threshold.
- I was not able to submit the NPS block. Clicking submit button did nothing for me. No errors on console.

ice9js:
> Based on the test plan, all worked as advertised!
> 
> Two things I noticed:
> 
> * In "Preview" mode, the NPS block did not show immediately, but needed to pass the threshold.
> * I was not able to submit the NPS block. Clicking submit button did nothing for me. No errors on console.

These were only implemented in #20 which wasn't merged into the main feature branch yet.

PR #22, opened 2021-01-19
Add the results section to the NPS block sidebar

ice9js:
This patch updates the sidebar of the NPS block to add the results section. It also removes the viewThreshold from the sidebar which will be implemented as part of the toolbar (#21).  
The 'title' input inside the block has also been removed in favor of the `title` input inside the results area in the sidebar. It's populated with the rating question by default.

# Testing

- Add a new NPS block and focus it. You should see the results section in the sidebar.
- The results should be disabled until you save the block to the platform by clicking the `save` button on the block.
- Once the block is saved, the `View results` button should redirect you to the results page.

### DIFF
--- a/client/blocks/nps/edit.js
+++ b/client/blocks/nps/edit.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import React, { useState } from 'react';
-import { pick, times } from 'lodash';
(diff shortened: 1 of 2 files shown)


digitalwaveride (approved):
All the things on the test plan work as promised!
🚀

PR #23, opened 2021-01-20
Add NPS Color settings

ice9js:
This patch adds the color settings for the NPS block.

The available settings right now are background, button and text color, but I think it'd probably make sense to add an additional setting for the button text color and either border color, or update the top border to always match the button color? What do you think @digitalwaveride?

In addition, I refactored `withFallbackStyles` which used to accept a component and a function to generate the fallback values but since we were pretty much using the same code for all our blocks, I decided it'd make more sense to build that right into the HoC and standardize the fallback value names across the components which should results in minimal overhead to implement this in future blocks.
There's no gains in optimizing the 'probing' component as it's removed as soon as the values have been obtained anyway. So we might just as well throw everything we need in there.

Screen Shot 2021-01-20 at 3 04 38 PM

# Testing

Because this PR is also affecting how fallback colors work for our other blocks, verify the applause, poll and vote block are still displayed with the correct default colors in both the editor and the published view.

For the NPS block, there should be a new **Block styling** section in the sidebar. Change the values and make sure the correct areas of the block are affected in the editor as well as on the post page.


digitalwaveride:
> The available settings right now are background, button and text color, but I think it'd probably make sense to add an additional setting for the button text color and either border color, or update the top border to always match the button color?

Yes, let's add a "button text color"!
Order of styling options in the side bar should be then:
    1 Background Color
    2 Text Color
    3 Button Color
    4 Button Text Color

Regarding the border color, in order to keep the styling simple, it's already long enough,  let's update the top border to always match the button color.

Same for the Submit button, that is currently seems not to be hooked to the button color. 

Also Text Color did not work for me:
image

This included the Preview View:

image


One note regarding the hover state for the buttons: On the other crowdsignal-forms blocks we defined the hover state for the buttons in an "inverse state", meaning if the button is:
bg: green
text: white

then the hover state is:
bg: white
text: green

digitalwaveride (approved):
Awesome Kuba!!!
Works! 🚀 🚀 🚀

PR #24, opened 2021-01-20
Add a notification/preview bar to the block

ice9js:
This patch adds a notification above the NPS block in the editor, informing the user the block is actually a popup, and offering a preview link which opens the post preview in a new tab.

The block will be 'grayed out' when unselected to further highlight this fact. The block always switches to the rating view when unselected.

Screen Shot 2021-01-20 at 6 19 08 PM

# Testing

Add an NPS block and verify there's a notification banner on top. Clicking the **Preview** button should open a new post preview.  
Check that the block gets 'grayed out' when unselected and that the rating view is shown even if feedback view was previously active.  
Clicking into the block again should go into the last active view.

### DIFF
--- a/client/blocks/nps/edit.js
+++ b/client/blocks/nps/edit.js
@@ -3,12 +3,15 @@
  */
 import React, { useState } from 'react';
 import { pick, tap, times } from 'lodash';
+import classnames from 'classnames';
 
 /**
  * WordPress dependencies
  */
+import { Icon, Notice } from '@wordpress/components';
 import { RichText } from '@wordpress/block-editor';
-import { dispatch } from '@wordpress/data';
+import { dispatch, withSelect } from '@wordpress/data';
+import { compose } from '@wordpress/compose';
 import { __ } from '@wordpress/i18n';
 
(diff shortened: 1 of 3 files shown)


digitalwaveride (changes requested):
Thanks a lot Kuba! Very nice!!!
 
Test Plan works as described 👍 , though I would have one little behaviour change request, and two little styling adjustments, which we also could fix later:

- When the user unselects the block, I like that it switches back to the rating view, but I am a bit surprised when I select it again and it sometimes switches to the latest state (Feedback tab), and sometimes not. 
I would suggest always to start with the Rating tab on select. Might be sometimes one click more for the user, but eliminates the "surprising element" of that interaction, which is more annoying.

- The padding of the block is a bit too wide, currently 48px, in the design files we have 32px.

image

- The border top of the block is currently 5px, in the design files we have a 10px.

And all icons (incl. the eye) are still dummy I guess...

digitalwaveride (approved):
Works!!!
And thanks for the styling adjustments 🙇

PR #25, opened 2021-01-21
Update default and placeholder values for the NPS block

ice9js:
This patch updates the block placeholders and adds default values for the fields that likely make much sense given the very specific nature of the block.  
It also enables editing of the feedback placeholder text.

Screen Shot 2021-01-21 at 3 31 20 PM

# Testing

- Create w new NPS block
- All the fields should be pre-populated with values from the designs.

### DIFF
--- a/client/blocks/nps/attributes.js
+++ b/client/blocks/nps/attributes.js
@@ -19,9 +19,19 @@ export default {
 	buttonTextColor: {
 		type: 'string',
 	},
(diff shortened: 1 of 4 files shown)


digitalwaveride (approved):
Nice!  🚀 🚢 

Seeing it in some themes with huge h3 font-sizes makes me now wonder because of the length of the question, if we need to adjust this in some way ... but for now let's roll with it!

PR #28, opened 2021-02-01
Inject 'source' prop on payload

CGastrell:
This PR adds a forced injection on NPS payload through the API Gateway

This PR depends on D56246-code

## Test instructions
Checkout and create a post with an NPS block. Save the NPS. Make sure the dependent backend diff is applied.
Check your Crowdsignal dashboard. The newly created NPS should show with only the Results link available.

### DIFF
--- a/includes/gateways/class-api-gateway.php
+++ b/includes/gateways/class-api-gateway.php
@@ -171,10 +171,15 @@ class Api_Gateway implements Api_Gateway_Interface {
 	 * @return Nps_Survey|WP_Error
 	 */
 	public function update_nps( Nps_Survey $survey ) {
+		$survey_array = $survey->to_array();
+
+		// Inject source attribute.
+		$survey_array['source'] = 'crowdsignal-forms';
+
 		$response = $this->perform_request(
 			'POST',
 			$survey->get_id() ? '/nps/' . $survey->get_id() : '/nps',
-			$survey->to_array()
+			$survey_array
 		);
 
 		if ( is_wp_error( $response ) ) {


includes/gateways/class-api-gateway.php
```diff
@@ -171,10 +171,15 @@ public function unarchive_poll( $id_to_unarchive ) {
 	 * @return Nps_Survey|WP_Error
 	 */
 	public function update_nps( Nps_Survey $survey ) {
+		$survey_array = $survey->to_array();
+
+		// Inject source attribute.
+		$survey_array['source'] = 'crowdsignal-forms';
```
ice9js, line 177: I wonder if this should maybe be part of the `NPS_Survey` class?
It's a constant value so we could inject it as inside the `to_array()` method.

I believe we'll need to add a `source_link` property on that class as well to be compatible with d56246-code?

ice9js (changes requested):
I think we might be missing the `source_link` prop :)

includes/gateways/class-api-gateway.php
CGastrell: > I wonder if this should maybe be part of the NPS_Survey class?

Wonder as well. While `source` and `source_link` are accepted payload props, they are not required and, actually, not customizable from the client side. I consider these props *hard coded* for a reason: we don't want client side messing with those other than through our automated way of setting them.

Having said that, I placed the `source` injection out of reach of any method, but I'm open to discuss better approaches for it.

> I believe we'll need to add a source_link property on that class as well to be compatible with d56246-code?

As talked earlier, knowing this will require some class|model modification, the changes on d56246-code are only paving the road so the props are handled properly. Next PR will/should have the needed changes/additions to send the props

ice9js: Well, thinking about it some more... this is very much the 'end of the line' before the request go out so it makes some sense in that way. Let's stick with it for now then.

ice9js (approved):
As we discussed on Slack, `source_link` will be part of a separate PR so no need to hold this one up any longer.

PR #29, opened 2021-02-02
Add/source link to payload

CGastrell:
This PR will append the `source_link`/`sourceLink` attribute to the payload sent to the API.

The `source_link` is taken from core selector `getPermalink`.

## Test instructions
Checkout and `make client`. Save an NPS block and confirm you can see the link on your dashboard under the column "Source" for the saved NPS survey.

### DIFF
--- a/client/blocks/nps/edit.js
+++ b/client/blocks/nps/edit.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import React, { useEffect, useState } from 'react';
-import { pick, tap, times } from 'lodash';
+import { times } from 'lodash';
 import classnames from 'classnames';
 
 /**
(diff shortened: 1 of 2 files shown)


client/blocks/nps/edit.js
```diff
@@ -57,11 +58,12 @@ const EditNpsBlock = ( props ) => {
 		try {
 			const { surveyId } = await updateNps(
 				tap(
-					pick( attributes, [
+					pick( { ...attributes, sourceLink }, [
```
ice9js, line 61: I'd say, either add it to the `data` object in the function below - similar to how the title fallback is set.
Or just rewrite the whole thing as:

```
const { feedbackQuestion, ratingQuestion, surveyId, title } = attributes;

updateNps( {
    feedbackQuestion,
    ratingQuestion,
    sourceLink,
    surveyId,
    title: title || ratingQuestion,
} );
```

The latter is probably the best for easy readability. I admit the `tap/pick` combo might've been a bit overkill.

includes/models/class-nps-survey.php
```diff
@@ -49,6 +49,14 @@ class Nps_Survey {
 	 */
 	private $feedback_question = '';
 
+	/**
+	 * Permalink URL where the NPS is published.
+	 *
+	 * @since [next-version-number]
```
ice9js, line 55: So that's 1.4? I like the idea with the placeholder but should we maybe document that somewhere to bulk-update these before a release? (That'd be out of scope for this PR).

includes/models/class-nps-survey.php
```diff
@@ -75,7 +84,8 @@ public static function from_block_attributes( $attributes ) {
 			$attributes['surveyId'],
 			$attributes['title'],
 			$attributes['ratingQuestion'],
-			$attributes['feedbackQuestion']
+			$attributes['feedbackQuestion'],
+			! empty( $attributes['sourceLink'] ) ? $attributes['sourceLink'] : ''
```
ice9js, line 88: Nit: I believe this could be simplified to `$attributes['sourceLink'] ?: ''`?

ice9js (changes requested):
Nice! Just picked at a few minor things :)

includes/models/class-nps-survey.php
CGastrell: The `prepare-release.sh` script takes care of it.

includes/models/class-nps-survey.php
CGastrell: I was just going with the linter:
image

ice9js: Didn't realize that was illegal 🤣

ice9js (approved):
🚢  it!

PR #30, opened 2021-02-03
Add autosaving capabilities for the NPS block

ice9js:
This patch adds autosave capabilities for the NPS block (and possibly others in the future). Solves c/wCM0MVuu-tr.

I accomplished this by creating a `withAutosave` higher order component which takes two arguments:

- An `onSave` handler, which should return a promise and accept two arguments:
  1. `attributes` which are the current attributes of the to be saved block.
  2. `setAttributes` in case the attributes need to be updated based on the logic inside the `onSave` handler.
- `watchedAttributes` which is a list of attribute keys to watch that should trigger the autosave behavior.

The onSave handler is debounced (I set it to a 1.5s default) which should give enough time to accumulate rapid changes into a larger request.
When the `onSave` handler fails, the HoC will retry the save operation until `RETRY_THRESHOLD` is met, after which a `saveError` prop will be set to `true` so our block can display an appropriate error message.

The HoC also inject a `forceSave` prop which can be used to allow the user to manually retry the save, or to save a new block to Crowdsignal.com as soon as it's created.

Screen Shot 2021-02-03 at 5 13 04 PM
Screen Shot 2021-02-03 at 6 00 02 PM

# Testing

- Create a new post and create a new NPS block. This block should appear in your Crowdsignal dashboard immediately, without the need to update any of its properties.
- Try changing the title or any of the questions, after a very short delay the changes should be reflected in your Crowdsignal dashboard without the need for any additional actions.
- Switch your browser to offline mode, and try making some more changes. An error banner should appear above the block after a few seconds.
  * Go back into online mode and click the _save_ button. The block should now be saved to Crowdsignal.com correctly.
  * Go back into online mode and make a change to the title or the block questions, the block should correctly autosave now.


digitalwaveride (changes requested):
The block seems to work, if I do a simple "publishing" flow. It also tracked as expected response data.

Though the "syncing" of any changes on the title or the question itself did not update on my dashboard, see the video. (I did just a basic "make client" ... maybe anything else that would have been necessary to set up for testing??

https://user-images.githubusercontent.com/5120357/106816759-218d1400-6676-11eb-90ba-1e092246c4fe.mov

ice9js:
After some discussion with @CGastrell yesterday on how this PR fits the changes from #29, I gave it some thought and decided a hook is probably a superior solution to a HoC in this case.

I managed to simplify the hook code quite a bit compared to its HoC counterpart, and the key difference with hooks is we're dealing with values directly rather than using keys to determine which values to watch. This allows us to easily work with both `props` and `attributes`, or any other variables we need.

The second argument of `useAutosave` accepts an object which will properties will be used to changes, and which will also be passed as a whole to the `onSave` callback. It is crucial that the callback relies on the argument, rather than using variables directly as compared to a regular `useCallback` hook as the latter would cause the function to be re-created on each render making the debounce problem much more difficult.

CGastrell (approved):
Works as advertised! Thanks for this Kuba!

A couple of thoughts/nits (not for this PR):

  - while attempting to save, there is no hint that something is going on. Watching/waiting for the block to fail on syncing, some 8-10 seconds seems a lot before actually telling the user "hey, something's wrong". Also true that this shouldn't happen at all. I wouldn't mind to show (not a spinner) some little text or something, but @digitalwaveride might have a better idea
  - once the sync has failed and the warning is shown: triggering a new sync by either hitting the "save" button or making a change "shakes" the content (as per the sudden disappearance of the warning)

digitalwaveride:
🚀 Worked!!! Awesomeness!!! Very cool Kuba!

PR #31, opened 2021-02-03
Support exclusive block

CGastrell:
This PR sets the `supports.multiple: false` for the NPS block.

## Test instructions
Checkout and `make client`. Insert an NPS block on a post, then try to insert another. The second time the NPS should not be available for insertion
image


digitalwaveride (approved):
image
Works for me! 🚀 

It would be kind of nice to give the user some feedback why the block is disabled in that case, but I would think that is an edge case either way. There is no logical reason to add two NPS blocks on a page.

PR #33, opened 2021-02-08
Add branding for NPS block

CGastrell:
This PR sets up branding for the NPS block:

  - Turns FooterBranding to a more independent component
  - Adds `message` option for the FooterBranding legend
  - Adds SignalWarning on both Sidebar and main block

### Screenshots
image
image

## Test instructions
Checkout and `make client`.

Test with free user over and under signal limit to see SignalWarning.
Verify that SignalWarning is not shown when within limits
Test with Business plan to verify no warning/branding is shown on editor nor public

### DIFF
--- a/assets/stylesheets/editor.scss
+++ b/assets/stylesheets/editor.scss
@@ -5,6 +5,7 @@
 @import "components/connect-to-crowdsignal/style.scss";
 @import "components/signal-warning/style.scss";
 @import "components/sidebar-promote/style.scss";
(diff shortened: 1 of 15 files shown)


digitalwaveride (approved):
Awesome Christian!
Works for me as advertised!

In a next step we need to adjust a bit the styling of the "banners" -> height, border and radius. There are some inconsistencies, with Gutenberg ... but maybe let's do this in a separate diff as it would impact all our blocks. Will ping you!

PR #34, opened 2021-02-09
Use Notice component for SignalWarning

CGastrell:
This PR:

  - introduces EditorNotice as a wrapper around Notice (`@wordpress/components/notice`)
  - turns SignalWarning into a of wrapper around EditorNotice for consistency
  - tweak error notice button style and text: now it reads "Retry"

It also turns some styles used inside the Notice into more generic styles so all our notices have a consisten spacing:
image

## Test instructions
Checkout and `make client`. Use a free account with exhausted signals to trigger the warning. See that it looks good and inline with any other notice present.


digitalwaveride:
Great Christian! Huge improvement!

Still not sure about the background styling... 
based on this ... https://developer.wordpress.org/block-editor/components/notice/

image


Will try to find out!

One little thing upfront, for the red warning banner, the color outline should be not blue I guess... does that color come from the component?

ice9js (changes requested):
Overall, nice :)

But I'm wondering, we should probably make this into a custom `EditorNotice` (`/client/components/editor-notice`) wrapper for Gutenberg's Notice. If we don't, it'll only help with consistency as long as we remember we need to replicate the markup. And we seem to be using it quite a lot.

CGastrell:
> Overall, nice :)
> 
> But I'm wondering, we should probably make this into a custom `EditorNotice` (`/client/components/editor-notice`) wrapper for Gutenberg's Notice. If we don't, it'll only help with consistency as long as we remember we need to replicate the markup. And we seem to be using it quite a lot.

Thought of it as well, just didn't want to refactor too much. On the other hand, it would just be a wrapper of Notice with a custom CSS class.

CGastrell:
Option 1, use `is-primary`
image

Option 2, use `is-destructive` (although it isn't actually destructive but *constructive*)
image

Option 3: create our own button class and apply it. Any thoughts @digitalwaveride ?

ice9js:
>it would just be a wrapper of Notice with a custom CSS class

Exactly - so if the styles belong to the respective block the class name should reflect that. If not, it should preferably be a separate component. I'd just like to avoid the case where we have some 'random' classes all over the place.

>Option 2, use is-destructive (although it isn't actually destructive but constructive)

⬆️

digitalwaveride:
Thanks Christian for playing with some options! 
Let's keep that for sure simple, and not try to make new classes.

> Option 2, use is-destructive (although it isn't actually destructive but constructive)
True, but I still would go with this "option 2" solution here, as we might give the class name's naming a little bit more meaning, as it needs to have in this particular context.

Maybe we could also optimize the button label ... instead of "save" calling it "retry". 
Then we also can shorten the notification text, and don't need the "click" sentence.

ice9js (changes requested):
Nice, my only complaint would be: the buttons should be right aligned.

Screen Shot 2021-02-12 at 8 03 00 PM

digitalwaveride (approved):
Agree with Kuba, otherwise, very nice! Thanks Christian!

ice9js (approved):
LGTM!

PR #35, opened 2021-02-09
Add status and scheduled disable

CGastrell:
This PR introduces the option to enable/disable the NPS popup or schedule disabling it after a certain date.

image
image
image

## Test instructions
Checkout and `make client`.

### Enabled
  - DateTime Picker should not be visible
  - Previewing the draft should always popup the NPS survey

### Disabled
  - NPS survey should not popup, not even on draft preview

### Disable after (should it be "Disabled after"?)
  - DateTime Picker should be visible when selecting this option
  - If date is in the past, a small hint should appear below the dropdown: "Currently disabled as date has passed"
  - Previewing draft should popup the NPS survey until the date/time has passed, not after

### DIFF
--- a/client/blocks/nps/sidebar.js
+++ b/client/blocks/nps/sidebar.js
@@ -129,15 +129,15 @@ const Sidebar = ( {
 					label={ __( 'Status', 'crowdsignal-forms' ) }
 					options={ [
 						{
-							label: __( 'Enabled', 'crowdsignal-forms' ),
+							label: __( 'Open', 'crowdsignal-forms' ),
 							value: NpsStatus.OPEN,
 						},
 						{
-							label: __( 'Disable after', 'crowdsignal-forms' ),
+							label: __( 'Closed after', 'crowdsignal-forms' ),
 							value: NpsStatus.CLOSED_AFTER,
 						},
 						{
-							label: __( 'Disabled', 'crowdsignal-forms' ),
+							label: __( 'Closed', 'crowdsignal-forms' ),
 							value: NpsStatus.CLOSED,
 						},
 					] }
@@ -147,7 +147,7 @@ const Sidebar = ( {
 						null !== attributes.closedAfterDateTime &&
 						new Date().toISOString() >
 							attributes.closedAfterDateTime
-							? 'Currently disabled as date has passed'
+							? 'Currently closed as date has passed'
 							: ''
 					}
 				/>
@@ -159,7 +159,7 @@ const Sidebar = ( {
 								new Date( attributes.closedAfterDateTime ) ) ||
 							new Date()
 						}
-						label={ __( 'Disable on', 'crowdsignal-forms' ) }
+						label={ __( 'Close on', 'crowdsignal-forms' ) }
 						onChange={ handleChangeCloseAfterDateTime }
 						is12Hour={ true }
 					/>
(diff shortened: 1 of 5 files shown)


digitalwaveride:
Awesome Christian! Thanks a lot for this!
Just one thing regarding the labeling... could we use the same terms as for the poll block to be consistent? Also regarding the dashboard etc.
 
image

includes/frontend/blocks/class-crowdsignal-forms-nps-block.php
```diff
@@ -168,6 +168,14 @@ private function attributes() {
 				'type'    => 'string',
 				'default' => 3,
 			),
+			'status'              => array(
```
ice9js, line 171: Nit: Just an observation I made... in terms of the block itself, it works just as well if we replace status with an `isOpen` or `isClosed` property. Then we could simplify the conditions to:

```
if isClosed:
    return;

// If a survey isn't closed but closedAfter is set, it implies the 'close after' state
if closedAfter && closedAfter < currentDate:
    return;

// display block
```

The advantage of this approach is it removes the need for the constants entirely which removes any overhead otherwise necessary to share the logic between the client and the server.

That said, if we're going to simplify it, we should probably do that before we launch the block as not to have to deal with supporting both ways for already saved blocks.

ice9js (approved):
I found a few nits but nothing critical. 👍

client/blocks/nps/sidebar.js
```diff
@@ -36,6 +39,13 @@ const Sidebar = ( {
 			[ attribute ]: value,
 		} );
 
+	const handleChangeStatus = ( status ) => setAttributes( { status } );
+
+	const handleChangeCloseAfterDateTime = ( closedAfterDateTime ) => {
+		const dateTime = new Date( closedAfterDateTime );
```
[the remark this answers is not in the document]
CGastrell: not sure, I went with the same implementation as in Polls, I just checked we got a consistent value

includes/frontend/blocks/class-crowdsignal-forms-nps-block.php
CGastrell: Well, we could have a nice function to return `isClosed` value based on the status and date. Kinda looks the same to me as a `isClosed` attribute could only be derived via a function/calculation you mention, but open to suggestions. I always go for the *raw data* approach and then derive values depending on context.

client/nps.js
```diff
@@ -23,6 +24,18 @@ window.addEventListener( 'load', () =>
 
 				element.removeAttribute( 'data-crowdsignal-nps' );
 
+				if ( NpsStatus.CLOSED === attributes.status ) {
```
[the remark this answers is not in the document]
CGastrell: Mmm... good point. Make it a PR? :)

includes/frontend/blocks/class-crowdsignal-forms-nps-block.php
ice9js: I am talking about simplifying the raw data though, by replacing the tri-state `status` with a binary `isClosed` value. This constrains the domain a bit which gets rid of some theoretical undefined states.

But like I said, it's nitpicking. I just can't resist the opportunity to write less code :D

PR #36, opened 2021-02-10
Add keywords and example

CGastrell:
This PR adds

  - some keywords for discovery on the block search
  - the example to be shown on the block search
  - the block icon

The example demanded an extra attribute `isExample` so we could prevent the triggering of the autosaving function.

image

Suggestions are welcome.

## Test instructions
Checkout and `make client`. Use the block search with any of the keywords. See the example renders on mouse over.


CGastrell:
The example needs to render a different block as it's creating a new NPS survey on every *hover*

ice9js (changes requested):
Just one more nit: can we update the classes so `is-inactive` isn't applied when `isExample === true` ?

ice9js (approved):
:shipit:

digitalwaveride:
Here some keyword suggestions:

net promoter score
nps
rating
feedback
customer satisfaction
review
survey
poll
form
Crowdsignal
loyalty
customer experience
CSAT

digitalwaveride:
And here a first suggestion of the description:

`Measure your Net Promoter Score! Collect feedback and track customer satisfaction over time.`

digitalwaveride:
Aaaannnd... and already an edit needed, sorry. 

I forgot that we need to adjust the title of the block slightly.

Title: 
`Measure NPS`

Description:
`Calculate your Net Promoter Score! Collect feedback and track customer satisfaction over time.`

image

digitalwaveride (approved):
YAY! 🚢  
Thanks Christian!

PR #37, opened 2021-02-12
Introduce new NPS block

CGastrell:
This PR merges back the feature branch into master for next release, which includes the new NPS block as well as some minor fixes to our current plugin version.

## Test instructions

Testing has been conducted for the individual bits in their respective PRs.
Check that the block looks and works like expected.


ice9js:
Let's go 🚀

PR #38, opened 2021-02-15
Add a submit view for the NPS block

ice9js:
This patch adds the submit view for the block along with the branding (which was previously only present in the editor).

Screen Shot 2021-02-15 at 4 44 11 PM

The view is shown after the user submits their feedback, unless the request failed in which case the dialog will just close.

# Testing

Create a new block and proceed to submitting a rating and feedback. You should see the new 'submit view' instead of the dialog disappearing.

### DIFF
--- a/client/blocks/nps/constants.js
+++ b/client/blocks/nps/constants.js
@@ -1,6 +1,7 @@
 export const views = {
 	RATING: 'rating',
 	FEEDBACK: 'feedback',
+	SUBMIT: 'submit',
 };
 
 export const NpsStatus = Object.freeze( {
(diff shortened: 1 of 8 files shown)


digitalwaveride:
Works as advertised!!! 
Though the branding shows twice for me... :

image
image

digitalwaveride (changes requested):
See comment above.

digitalwaveride:
Double branding is gone! Perfect! Thanks Kuba!

Playing with different themes though, realized we need to give the branding some more "fixed" margin-top, and also the font-size and placement of the branding is off.

image

image

The main problem in the case above is that the branding text has below more margin than above. That's why it looks weird.

Besides that, the branding text is too big. The font-size of the branding text should be fixed 12px, independent of the theme styling. The padding below the branding text just 24px, the margin above the branding text 40px.

image

digitalwaveride (approved):
🚀  Perfect! Thanks a ton for the fixes!

PR #39, opened 2021-02-15
Update the notification bar to display the correct view threshold

ice9js:
A small change that will update the NPS block notification when the view threshold is changed to reflect the current value.

# Testing

- Create an NPS block - by default the notification should say the block will be displayed after 3 page views.
- Try changing the view threshold.
- Verify the notification is updated to display the current value.


digitalwaveride (approved):
Perfect! Works 🚀 ! Thanks Kuba!

PR #40, opened 2021-02-15
Tweak NPS block styles

ice9js:
This block addresses some of the remaining styling issues with the NPS block:

- Updated the rating scale labels to use system fonts instead of themes'
- Improved the inconsistency issue with the submit button on the feedback view between the editor and the page
- Adjusted paddings/margins and rating buttons for mobile view (<670px)
- Increased the font size for the feedback textarea
- Added a fixed `margin-top` for the feedback textarea

# Testing

Just stylistic changes. Run the build and verify that things look as expected :)

### DIFF
--- a/client/blocks/nps/util.js
+++ b/client/blocks/nps/util.js
@@ -11,6 +11,7 @@ export const getStyleVars = ( attributes, fallbackStyles ) =>
 			buttonTextColor:
 				attributes.buttonTextColor || fallbackStyles.textColorInverted,
 			textColor: attributes.textColor || fallbackStyles.textColor,
+			textSize: fallbackStyles.textSize,
 		},
 		( _, key ) => `--crowdsignal-forms-${ kebabCase( key ) }`
 	);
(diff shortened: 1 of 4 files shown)


digitalwaveride:
Kuba thanks so much for all the adjustments! Looking super good!


One styling change that I am not seeing for weird reasons... and three little touches:

### Input field text font-size is still small:

image

On mobile view the font is increased though:
image


### The margin above input field... 
still feels dense for the published view, increasing it to "margin-top: 32px;" would make sense... 

image


### One additional fine-tuning thing:
The padding above the labels of the scale feels much better on the editor than on the published view as well... we should increase it to margin-top: 32px there as well.

image


### One little glitch

For some themes (exp. twentytwenty the buttons get in the published view on hover an underline)

image

digitalwaveride:
Thanks Kuba! Awesome! All fixed, the only thing I am still seeing is the "underlined numbers" in the rating buttons in the published view of the Twentytwenty theme:

image

digitalwaveride (approved):
🚢 🚢 🚢 
Perfect!!! Thanks Kuba!

PR #42, opened 2021-02-18
PoC: Use native fetch for WPCOM

CGastrell:
Still a PoC. Works against D57291-code 

This PR implements a native fetch call to request the API. 

It adds a new endpoint `/nonce` on the NPS controller to be available for a preflight request. The fetch does the roundtrip and then makes the actual request with the newly acquired nonce.

The fetch mechanism is here for testing purposes while we do a call for testing. The mechanism should become a separate module to be used externally.

## Test instructions
Apply D57291-code and used a sandboxed site.

Checkout this PR and run `make client`.

Add an NPS block to a post, then visit it to check functionality. Then open the post with an incognito window and check functionality again.

Both times the requests should make it through and be viewable on the results page of the NPS

### DIFF
--- a/client/data/nps/index.js
+++ b/client/data/nps/index.js
@@ -21,8 +21,32 @@ export const updateNps = ( data ) =>
 		} )
 	);
 
-export const updateNpsResponse = ( surveyId, data ) =>
-	withRequestTimeout(
+export const updateNpsResponse = ( surveyId, data ) => {
+	console.log( window.csFormsSetup );
+	if ( window.csFormsSetup || window.csFormsSetup._isWpcom ) {
+		return window
+			.fetch(
+				'https://public-api.wordpress.com/crowdsignal-forms/v1/sites/' +
+					window.csFormsSetup._currentSiteId +
+					'/nps/' +
+					surveyId +
+					'/response?_locale=user',
+				{
+					headers: {
+						accept: 'application/json, */*;q=0.1',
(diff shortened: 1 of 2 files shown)


client/data/nps/index.js
```diff
@@ -21,8 +21,32 @@ export const updateNps = ( data ) =>
 		} )
 	);
 
-export const updateNpsResponse = ( surveyId, data ) =>
-	withRequestTimeout(
+export const updateNpsResponse = ( surveyId, data ) => {
+	console.log( window.csFormsSetup );
+	if ( window.csFormsSetup || window.csFormsSetup._isWpcom ) {
```
CGastrell, line 26: that should have been a `&&`

client/data/nps/index.js
```diff
@@ -21,8 +21,32 @@ export const updateNps = ( data ) =>
...
+						'x-wp-nonce': window.csFormsSetup._nonce,
+					},
+					referrer: window.csFormsSetup._siteUrl,
+					referrerPolicy: 'strict-origin-when-cross-origin',
+					body: JSON.stringify( data ),
+					method: 'POST',
+					mode: 'cors',
+				}
+			)
+			.then( ( response ) => response.json() );
```
CGastrell, line 47: there is no error handler, I was hoping we can raise it and handle it on the caller

CGastrell:
This was addressed on other PRs/Issues

PR #43, opened 2021-02-19
Add a lightweight fetch wrapper for use on public pages

ice9js:
This patch supersedes #42 and implements a custom lightweight fetch wrapper to be used on the post page instead of `wp.apiFetch`. This way we don't need to add environment specific logic to the plugin itself.

The main issue with using `wp.apiFetch` on public pages is that it's essentially a singleton and it's optimized to be used within the editor context which can lead to unexpected issues when attempting to use it on the front-end of the site, especially if there's other plugins or scripts adding their own behaviors.

For more context, see pabtAt-1cn-p2.

`client/apifetch.js` is essentially just `window.fetch` with the addition of support for middlewares.  
In order for the middleware to be usable - in case the request needs to be altered for your WP install - the module is exposed globally on `crowdsignalForms.apiFetch`.

# Testing

Run `make client` and verify that creating NPS blocks and submitting responses still works as expected on self-hosted WP installs.


digitalwaveride (approved):
🚀 

Things still working for me! Responses are recorded, NPS block behaves as expected.

client/apifetch.js
```diff
@@ -0,0 +1,115 @@
...
+ * Appends a middleware to apiFetch
+ *
+ * @param  {Function} middleware Middleware function
+ */
+apiFetch.use = ( middleware ) => middlewareRegistry.push( middleware );
+
+/**
+ * Prefix the request URLs with '/wp-json'
+ */
+apiFetch.use( ( options, next ) => {
```
CGastrell, line 49: If this middleware is hardcoded here, how would we go adding another one for a different env where the path is completely different?

Maybe I'm still not grasping the concept right, but trying to add another middleware to manipulate the path seems hacky at this point instead of flexible?

client/apifetch.js
```diff
@@ -0,0 +1,115 @@
...
+		return next( options );
+	}
+
+	return next( {
+		credentials: 'same-origin',
+		mode: 'same-origin',
+		...options,
+		headers: {
+			'X-WP-Nonce': window._crowdsignalFormsWpNonce,
+			...options.headers,
```
CGastrell, line 110: Shouldn't the spread be inverted? Spreading `options` and `options.headers` **after** other properties have been added would overwrite any of those properties with the same key/name

CGastrell (approved):
Overall this seems great and, specially, tidy and modular. 

I'm concern regarding the hardcoded middlewares as this reproduces the same problems with `wp.apiFetch` where adding a middleware means any hardcoded middleware already took place, hence, maybe invalidating the one you're trying to add.

I left an inline comment with the most basic but tangible concern regarding the `path` manipulation: adding a middleware there is already too late as the path has already been changed and the original one is not an option anymore (turning the `wp-json` path into a public-api one would now require parsing/matching the path)

Maybe invert the middlewareRegistry before running it?

Anyways, let's keep at it. Great work!

client/apifetch.js
ice9js: This is actually correct, the idea is, if you went to the lengths to manually override `X-WP-Nonce` it should override this middleware.

ice9js:
I made some tweaks: put the middlewares in a separate module and added a `disable` method.  
If you'd like to disable a middleware from __vanilla JS__ you can now do it like that:

```
crowdsignalForms.apiFetch.disable( crowdsignalForms.apiFetch.middleware.formatURL );
```

CGastrell:
> I made some tweaks: put the middlewares in a separate module and added a `disable` method.
> If you'd like to disable a middleware from **vanilla JS** you can now do it like that:
> 
> ```
> crowdsignalForms.apiFetch.disable( crowdsignalForms.apiFetch.middleware.formatURL );
> ```

Cool! Although, for once, I'd follow the way `wp.apiFetch` does it: provide the default middlewares (the ones that do what we need at Crowdsignal) and let them be applied:
```
crowdsignalForms.apiFetch.use( crowdsignalForms.middleware.selfHostedPath );
crowdsignalForms.apiFetch.use( crowdsignalForms.middleware.respondJson );
...
```
That way, you always have the apiFetch in a *raw* state, and you add the middlewares you need.

But, as I said, these are all things we can improve or decide later, don't hesitate to merge this :)

ice9js:
>Cool! Although, for once, I'd follow the way wp.apiFetch does it: provide the default middlewares (the ones that do what we need at Crowdsignal) and let them be applied.

The drawback to this would be that even if you're just installing it on a self-hosted site, you now have to go to the trouble of adding your own JS dependency to actually configure this. In case of `wp.apiFetch`, I believe they do a lot of what we use middlewares for as part of the custom fetch handler directly which is causing its own issues.

PR #44, opened 2021-02-22
Add credential-less nonce create/verify mechanism

CGastrell:
This PR makes the nonce be created without any user session data. This allows for consistent nonce validation which can't be done between the render request and the REST API request.

## Test instructions
Checkout and `make client`. Visit a post with an NPS block and submit both rating and feedback. You shouldn't get a 403/500 response.

### DIFF
--- a/includes/frontend/blocks/class-crowdsignal-forms-nps-block.php
+++ b/includes/frontend/blocks/class-crowdsignal-forms-nps-block.php
@@ -80,7 +80,13 @@ class Crowdsignal_Forms_Nps_Block extends Crowdsignal_Forms_Block {
 
 		$attributes['hideBranding'] = $this->should_hide_branding();
 		$attributes['isPreview']    = is_preview();
-		$attributes['nonce']        = wp_create_nonce( self::NONCE );
+
+		// store the logged in user ID.
+		$actual_user_id = wp_get_current_user()->ID;
+		// temporarily set user ID to 0 so we create a consistent nonce.
+		wp_set_current_user( 0 );
+		$attributes['nonce'] = wp_create_nonce( self::NONCE );
+		wp_set_current_user( $actual_user_id );
 
 		return sprintf(
 			'<div class="crowdsignal-nps-wrapper" data-crowdsignal-nps="%s"></div>',
--- a/includes/frontend/blocks/class-crowdsignal-forms-nps-block.php
+++ b/includes/frontend/blocks/class-crowdsignal-forms-nps-block.php
@@ -184,12 +184,31 @@ class Crowdsignal_Forms_Nps_Block extends Crowdsignal_Forms_Block {
 	 * The nonce creation is first attempted through crowdsignal_forms_nps_nonce filter.
 	 *
 	 * @since [next-version-number]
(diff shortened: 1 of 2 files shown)


includes/frontend/blocks/class-crowdsignal-forms-nps-block.php
```diff
@@ -80,7 +80,13 @@ public function render( $attributes ) {
 
 		$attributes['hideBranding'] = $this->should_hide_branding();
 		$attributes['isPreview']    = is_preview();
-		$attributes['nonce']        = wp_create_nonce( self::NONCE );
+
+		// store the logged in user ID.
+		$actual_user_id = wp_get_current_user()->ID;
+		// temporarily set user ID to 0 so we create a consistent nonce.
+		wp_set_current_user( 0 );
```
ice9js, line 87: So, I still stand by my opinion this isn't really the place for this - considering our current solution works perfectly fine on WordPress.org sites.

What we should to is create a filter, say `crowdsignal_forms_api_nonce` which should default to `wp_create_nonce`.  The validation should stay the way it is.

This filter can then be used to override the nonce value when required, while keeping the optimal/default behavior for basic WP sites.

ice9js (changes requested):
I still stand by my opinion that we should probably use filters to hook into this rather than hardcoding it right inside the plugin.

includes/frontend/blocks/class-crowdsignal-forms-nps-block.php
```diff
@@ -178,4 +178,18 @@ private function attributes() {
...
+
+	/**
+	 * Returns a nonce based on the NONCE.
+	 * The nonce creation is first attempted through crowdsignal_forms_nps_nonce filter.
+	 *
+	 * @since [next-version-number]
+	 */
+	private function create_nonce() {
+		$nonce = apply_filters( 'crowdsignal_forms_nps_nonce', self::NONCE );
+		if ( empty( $nonce ) ) {
```
ice9js, line 190: I think this is an error. If no filters are given than `$nonce = self::NONCE`; which kind of breaks.

includes/rest-api/controllers/class-nps-controller.php
```diff
@@ -116,8 +117,10 @@ public function upsert_nps_response( \WP_REST_Request $request ) {
 		$data      = $request->get_json_params();
 		$survey_id = $request->get_param( 'survey_id' );
 
+		$verifies = wp_verify_nonce( $data['nonce'], Crowdsignal_Forms_Nps_Block::NONCE );
```
ice9js, line 120: Also here, we should probably have a custom `verify_nonce` method which basically does `$nonce === $this->create_nonce()`. Just to make sure if any filters were applied they also apply for the verification.

ice9js (changes requested):
Nice, I like this approach 👍  Just a few small bits that need ironing out.

Overall, I'm not sure if this is the best place for this code but it's a good start. Just something to think about if you get any better ideas in the future :)

ice9js (approved):
Works for me now :shipit:

PR #48, opened 2021-02-24
Remove console debugging message

### LINKED ISSUE #27: window.p2tenberg is deprecated
`window.p2tenberg` is deprecated and at some point will be removed. If the Crowdsignal function `isP2tenberg` is still need then you can use `window.p2editor` instead.

It's unlikely the removal will happen soon, so there's no rush!

CGastrell:
This PR removes the console message used for debugging.

Also addresses #27 by searching for either `p2tenberg` or `p2editor` on `window` object.

Closes #27

### DIFF
--- a/client/components/with-poll-base/index.js
+++ b/client/components/with-poll-base/index.js
@@ -21,7 +21,7 @@ import usePollDuplicateCleaner from 'components/use-poll-duplicate-cleaner';
 
 startSubscriptions();
 
-const isP2tenberg = () => 'p2tenberg' in window;
+const isP2tenberg = () => 'p2tenberg' in window || 'p2editor' in window;
 
 const withPollBase = ( Element ) => {
 	return ( props ) => {
--- a/client/nps.js
+++ b/client/nps.js
@@ -44,12 +44,6 @@ window.addEventListener( 'load', () =>
 
 					window.localStorage.setItem( key, viewCount );
 
-					// eslint-disable-next-line no-console
-					console.log(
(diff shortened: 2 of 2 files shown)


client/components/with-poll-base/index.js
```diff
@@ -21,7 +21,7 @@ import usePollDuplicateCleaner from 'components/use-poll-duplicate-cleaner';
 
 startSubscriptions();
 
-const isP2tenberg = () => 'p2tenberg' in window;
+const isP2tenberg = () => 'p2tenberg' in window || 'p2editor' in window;
```
ice9js, line 24: Is this supposed to be in this patch?

client/nps.js
```diff
@@ -44,12 +44,6 @@ window.addEventListener( 'load', () =>
 
 					window.localStorage.setItem( key, viewCount );
 
-					// eslint-disable-next-line no-console
```
ice9js, line 47: Should we also just remove the command? We should do it for the release anyway and the preview allows you to to trigger it manually anyway.

ice9js (changes requested):

client/components/with-poll-base/index.js
CGastrell: Yeah, it's that old issue/suggestion from @johngodley I thought I could throw it here and not make another PR for it, but happy to decouple it if it seems messy

ice9js (approved):
:LGTM:

PR #49, opened 2021-02-24
Fix an issue with threshold icon margin in the toolbar

ice9js:
This fixes an issue with the icon margin styles for the toolbar icon for setting the NPS threshold.
The unwanted margin is caused by the popover adding another `<span>` next to the icon.

Before:

Screen Shot 2021-02-24 at 9 37 45 AM

After:

Screen Shot 2021-02-24 at 9 37 30 AM

# Testing

This doesn't actually affect the default Gutenberg setup but shouldn't break it either ;)

### DIFF
--- a/client/blocks/nps/edit.scss
+++ b/client/blocks/nps/edit.scss
@@ -39,7 +39,7 @@
 	}
 
 	&.components-button.has-icon .dashicon {
-		margin-right: 0;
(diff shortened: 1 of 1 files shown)


digitalwaveride (approved):
Thanks so much for the fast fix Kuba!!!! 🚀

PR #50, opened 2021-02-24
Use is singular instead of is_single

CGastrell:
This PR addresses a report that came in during tests where the block would not be rendered at all. The call on `is_single()` is returning false, which prevents the block from rendering.

As per [codex recommendation](https://developer.wordpress.org/reference/functions/is_single/#more-information), this PR changes `is_single()` call for `is_singular()`

## Test instructions
Checkout on sandbox and run `make client`. Be sure to sandbox ParKff-9N-p2 and visit the page where the NPS was inserted: ParKff-9W-p2
After 3 visits, the NPS should popup.


ice9js (approved):
👍

PR #54, opened 2021-03-05
Turn html and reusable options to false

CGastrell:
This PR changes the NPS `supports` options `html` and `reusable` to false. The NPS block is not prepared to be manually edited nor turned into a reusable one.

## Test instructions
Checkout and `make client`. Try to edit the block on HTML mode, you shouldn't be able to. Try to make an NPS block into a reusable one, you shouldn't be able to do so.


digitalwaveride (approved):
Thanks Christian!

PR #55, opened 2021-03-08
Use RichText component on frontend

CGastrell:
This PR addresses the issue seen on CS Forms blocks which allow for multiple line paragraphs but then render a `<br>` on frontend.

Applying this PR will allow line breaks on:

  - NPS Rating question
  - NPS Feedback prompt
  - Poll question (and preview while disabled)
  - Poll note (and preview while disabled)

and prevent line breaks on:

  - NPS low/high rating labels
  - NPS Feedback submit button
  - Poll submit button

## Test instructions

**NOTE:** we are importing a `block-editor` component on the visitor frontend. While this is not optimal, verify the `RichText` component can be seen already loading both on WP.com and self hosted environments before building this PR.

Checkout and run `NODE_ENV=production make client`. Verify the conditions described above.

### DIFF
--- a/client/blocks/nps/edit.js
+++ b/client/blocks/nps/edit.js
@@ -224,6 +224,8 @@ const EditNpsBlock = ( props ) => {
 								) }
 								value={ attributes.lowRatingLabel }
 								allowedFormats={ [] }
+								multiline={ false }
+								disableLineBreaks={ true }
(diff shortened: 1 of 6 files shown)


ice9js (approved):
🚢 🇮🇹

PR #56, opened 2021-03-09
Change copy edits

CGastrell:
This PR changes labels/texts on the editor:

  - "Survey Status" instead of "Status" on the Sidebar -> Settings section
  - "Show this block after __ visits": use double underscore as a placeholder for the threshold before showing the popup

## Test instructions
Checkout and run `make client`. Check the above mentioned texts.


digitalwaveride (approved):
Thanks Christian! 👍

PR #58, opened 2021-03-16
Use optimistic requests for submitting NPS block results

ice9js:
This patch updates the submission flow for the NPS blocks so requests aren't blocking for the next steps.
Because of this update, we'll now be giving every user the complete flow regardless if they requests go out or results in errors - previously we'd just hide the dialog at any point in the process if something goes wrong.

# Testing

Create an NPS block and try submitting a response. When you submit rating or feedback you should be taken to the next step immediately.
Verify your results show up in the dashboard.

### DIFF
--- a/client/components/nps/feedback.js
+++ b/client/components/nps/feedback.js
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import React, { useState } from 'react';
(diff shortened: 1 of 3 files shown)


CGastrell (approved):
Works as advertised! :shipit:

PR #59, opened 2021-03-17
Update dependency to node-sass@5.x.x

CGastrell:
This PR updates the `node-sass` dependency version and fixes incompatibility with Linux 64bit environments

On a second thought, this might be useless until we update the calypso-build dependency as well (node-sass@4.x is a direct dependency of it)

## Test instructions
Checkout update/node-sass-dependency-v5 and run `make client` on a Linux 64 env. The build should complete without errors


CGastrell:
Giving up for now, until we need to support Linux64

PR #60, opened 2021-03-17
Update the NPS preview button to behave identically to the editor's one

ice9js:
This patch updates the preview button on the NPS block to behave the same way as the editor's preview button.  

Previously it was prone to issues when it'd redirect you to the preview of a yet unsaved post or show you an outdated version. With this upgrade you should get the same 'generating preview' screen and should always see the current version of the post.

In order to get this to work in the least invasive manner and without duplicating core's functionality I had to tweak `EditorNotice` component a little.  
With the tweak I made, it'll only put the first `child` into the `crowdsignal-forms__editor-notice-text` node and append the rest behind it. I believe this should actually be good in the long run as it now allows us to append custom elements or buttons without breaking the layout. If we ever need to fit more nodes into the text node we could just use an additional wrapper for that.

# Testing

Verify that the preview button triggers a save if the post hasn't been saved yet and that the preview always reflects the latest changes made in the editor.

### DIFF
--- a/client/blocks/nps/edit.js
+++ b/client/blocks/nps/edit.js
@@ -39,7 +39,6 @@ const EditNpsBlock = ( props ) => {
 		clientId,
 		fallbackStyles,
 		isSelected,
-		postPreviewLink,
 		setAttributes,
 		renderStyleProbe,
 		sourceLink,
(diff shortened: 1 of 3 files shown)


CGastrell:
I'm getting an error when attempting to create an NPS block with this PR:
image

I want to confirm you are not getting this @ice9js  before rejecting the PR

ice9js:
Hmm 🤔  This looks like it could be a Gutenberg issue. I haven't been able to replicate it yet but even looking at the error... we're not using any hooks within `EditorNotice` and leaving the new preview button as the sole suspect.

This would also explain a potential difference in our testing environment and why I'm not able to replicate it if we're using a different version of WP/Gutenberg.

CGastrell (approved):
Nice! I removed some pesky plugins and it works as expected!

PR #61, opened 2021-03-17
Use 'crowdsignal_forms_translations_path' filter to load translations

CGastrell:
This PR adds a filter to the path where translations are loaded from through the "crowdsignal_forms_translations_path" filter, in case the user wants to store/load translations from a different path.

## Test instructions
Checkout and load the editor, the filter currently defaults to the `/languages` dir.


ice9js (approved):
👍

PR #71, opened 2021-03-25
Feedback block setup

ice9js:
Set up the skeleton for building the feedback block.

# Testing

After patching `make client` should build all the necessary assets for the feedback block and it should become available within the editor.


digitalwaveride (approved):
Thanks Kuba!

PR #72, opened 2021-03-25
[PoC]: Floating feedback block

ice9js:
This patch is an initial stab at making a 'floating' block. There's still a lot of fine tuning and cleanup to do within the block itself, but I'm fairly happy with the `withFixedPosition` HoC responsible for some of the magic.

I noticed that the best way to move the toolbar is to move the entire block container rather than the block itself. It took some digging but turns out we can actually modify the wrapper by using a filter and a higher order component.

The downside is, that HoC cannot pass anything down to the block itself which means even though we can set a fixed position we wouldn't be able to update it. So naturally state management comes to mind but after thinking for a bit I came up with a slightly simpler idea.
Since what we're trying to do is so self contained, we can store references to components internal `set...` state functions outside of the component, and then have another HoC that injects these references directly into the block. 

The nice thing about this approach is the entire mechanism is fully self-contained within the two HoC's and doesn't require any additional setup from the block - making it actually quite portable.

As for how the positioning is handled within the block - this will probably change a bit, especially when we add the option for different button styles which would actually affect the behavior slightly as well.

This patch includes the changes from #71 for easy testing - I'll just rebase it before merging.

# Testing

When you add a `feedback` block to your page, it should appear in the lower right corner by default.
Clicking on the 'empty' button in the toolbar should display a menu of six locations for the block. Selecting any of them should move the block along with the toolbar to the desired location.

When the block is positioned on the left or the right, it should reposition itself correctly whenever the sidebar or full screen mode are toggled.


digitalwaveride (approved):
Quod erat demonstrandum.

Total awesomeness! Super excited about this! Great job Kuba!

CGastrell:
I'm getting a fatal from JS, but I'm starting to think there must be something wrong with my setup.

image

This time I don't have the Gutenberg plugin enabled (only the default installation one), but if both you and @digitalwaveride were able to run it, I'll leave it to your consideration.

PR #73, opened 2021-03-25
Add Github action for nightly builds

CGastrell:
This PR adds (actually just polishes it) a workflow to manually trigger a build.

When triggered, the action allows to type a branch name to be built (in case you need to build something else than master branch). **Do not confuse with the "from" branch** (first dropdown), which just lets you pick a different version of the action itself (for example, when you're editing the action on a branch, you can run the action from that particular branch to see how your changes work)
image

When the action ends successfully, you should be able to see the "Development Build" release as a pre-release:
image
and, most importantly, the archived version of the nightly build should be available for download among the assets.

## Test instructions
Go the repo actions: https://github.com/Automattic/crowdsignal-forms/actions
Select the "Create nightly build release" workflow
image
Use the "Run workflow" dropdown. Select "debug/action-steps" as the "Use workflow from" and target "test/get-rid-of-docker-step-for-pot-file" (this is the only branch that can be built right now as it doesn't use the docker to generate the .pot file).
Click again on the "Create nightly build release" to see the workflow triggered, it will show with a yellow dot indicating that it's still running. Wait for it to finish and then head to the releases section: https://github.com/Automattic/crowdsignal-forms/releases
You should see the "nightly_build" release just updated with the workflow you triggered

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -34,6 +34,7 @@ jobs:
       - name: Install
         run: |
           npm install
+          composer update
           composer install
 
       #
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -27,6 +27,11 @@ jobs:
         with:
           php-version: '7.4'
 
+      - name: Setup Node version
+        uses: actions/setup-node@v2
+        with:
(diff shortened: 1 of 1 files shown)


ice9js (approved):
LGTM!

PR #75, opened 2021-04-01
implement feedback question view

ice9js:
This patch implements the feedback question view as well as the basic color settings to control it. You should be able to edit any text/color of the popup.

Solves c/qA8oB2p8-tr and c/8dZT9xDQ-tr.

Screen Shot 2021-04-01 at 10 52 06 PM

I did try to improve the positioning inside the editor and the page a little - the corners should work ok now but the vertically centered one is still a bit wonky. I have a feeling CSS is probably the wrong approach here so I'll need to add a JS solution and do things a little different.  
For the editor, I'm almost satisfied except the script should take into account the popup's height when using vertically centered position. For the page, the idea would be to position the 'trigger' dynamically (so we can more easily accommodate different sizes in the future) and then have the popover be an actual popover placed relatively to the trigger rather than the page. The editor and the page have unique behaviors but they also don't share components which is why this approach actually makes sense.

# Testing

Apply the patch and add a feedback block to your post. You should be able to edit the header, feedback and email placeholders, and the submit button label.  
You should also be able to edit the block colors by opening the "Color Settings" in the sidebar.

Save the post and open the preview - make sure the block reflects your selected settings.

### DIFF
--- a/client/blocks/feedback/attributes.js
+++ b/client/blocks/feedback/attributes.js
@@ -25,7 +25,10 @@ export default {
 	},
 	feedbackPlaceholder: {
 		type: 'string',
-		default: __( 'Please let us know how we can do better...', 'crowdsignal-forms' ),
+		default: __(
+			'Please let us know how we can do better…',
+			'crowdsignal-forms'
+		),
 	},
 	header: {
 		type: 'string',
--- a/client/blocks/feedback/attributes.js
+++ b/client/blocks/feedback/attributes.js
@@ -1,20 +1,61 @@
 /**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
  * Note: Any changes made to the attributes definition need to be duplicated in
  *       Crowdsignal_Forms\Frontend\Blocks\Crowdsignal_Forms_Nps_Block::attributes()
  *       inside includes/frontend/blocks/class-crowdsignal-forms-nps-block.php.
  */
 
 export default {
+	backgroundColor: {
+		type: 'string',
+	},
+	buttonColor: {
+		type: 'string',
+	},
+	buttonTextColor: {
+		type: 'string',
+	},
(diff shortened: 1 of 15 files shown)


CGastrell:
Make lint happy

CGastrell (approved):
Inputs and styles are responding as advertised.

While positioning still needs improvement (or a decision for some more "fixed" rules/behavior) I see this close to the expectations.

When aligning the "trigger" to the `center` right or left, I would have expected that the popup would also be centered in relation to the trigger, yet it rendered as if it was anchored from the corner:

Expected on `center-right` alignment:
image

Current behavior:
image

digitalwaveride:
So cool Kuba!
Very very nice!
A few things I noticed:

- Block Styling seems to work in the editor, though they were not displayed in the preview ... the button color not responding to overwrite?:
Editor:
image

Preview:
image

- There is also something going on with the emoji styles, but not sure if we have any impact on this?

- The placeholders are not "looking" like placeholders, they would need some "greyish" styling.

- The input fields need some fine-tuning regarding "height" ...

- The modal window needs a transparent dark background filter, to reduce noise from background elements. Under some circumstances the modal could otherwise look like a "rendering" error or overall it could look very noisy.

ice9js:
As we discussed:

>There is also something going on with the emoji styles, but not sure if we have any impact on this?

Seems to be a theme related thing.

>The placeholders are not "looking" like placeholders, they would need some "greyish" styling.
>The input fields need some fine-tuning regarding "height" ...

Will add those to the backlog.

>The modal window needs a transparent dark background filter, to reduce noise from background elements. Under some circumstances the modal could otherwise look like a "rendering" error or overall it could look very noisy.

This could be a bit tricky but will try to figure something out later.
Overall, I'm going to merge this into the feature branch as is with the button and linter issues fixed to unblock upcoming things and address the smaller issues separately :)

PR #76, opened 2021-04-06
Add Feedback API functionality

CGastrell:
This PR depends on #75 

This PR adds all the backend files, methods and registration for the new Feedback block to work against the API `/feedback` endpoint.

## Test instructions
Checkout and `make client`.

  - Add a Feedback block to a post and publish it
  - Visit your Crowdsignal dashboard to see the newly created survey is shown
  - Visit the post where you published the Feedback block, submit some feedback
  - Visit the results page and check your submitted feedback can be seen


includes/rest-api/controllers/class-feedback-controller.php
```diff
@@ -0,0 +1,189 @@
...
+		$result = Crowdsignal_Forms::instance()->get_api_gateway()->update_feedback_response(
+			$survey_id,
+			$data
+		);
+
+		if ( is_wp_error( $result ) ) {
+			return $result;
+		}
+
+		$result['checksum'] = $this->get_response_checksum( $result['r'], $data['nonce'] );
```
ice9js, line 134: I believe we submit feedback only once so we don't actually need the checksum to make sure both responses are coming from the same source.

includes/rest-api/controllers/class-feedback-controller.php
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	/**
+	 * Creates a checksum hash for a response ID and nonce combination.
+	 *
+	 * @param  string $response_id Response ID.
+	 * @param  string $nonce       Nonce.
+	 * @return string
+	 */
+	private function get_response_checksum( $response_id, $nonce ) {
```
ice9js, line 186: This function could be deleted here.

ice9js (changes requested):
Nice @CGastrell, thanks! The only nitpick I have is to remove the checksum logic from the controller - I believe we don't need that for surveys (and it actually doesn't seem to be used anywhere anyway?).

ice9js (approved):
Thanks! :shipit:

PR #77, opened 2021-04-07
Fix signal warning layout

CGastrell:
This PR addresses an issue when the signal warning is rendered. The problem mostly happens because we use [Notice](https://github.com/WordPress/gutenberg/tree/trunk/packages/components/src/notice) default behavior (providing action buttons through an array) mixed with our need to use PostPreviewButton on one of the notices, which has its own behavior to build and open the preview URL.

The changes include a new prop for our EditorNotice: `componentActions` which behaves like the Notice's `actions` prop but assumes the items of the array to be components instead of objects to compose a button from.

## Test instructions
Confirm the issue before testing: when linked to a free Crowdsignal account with signals exhausted, the signal warning looks like this:
image

Checkout and `make client`. See that the notice renders properly:
image

### DIFF
--- a/client/blocks/nps/edit.js
+++ b/client/blocks/nps/edit.js
@@ -160,7 +160,22 @@ const EditNpsBlock = ( props ) => {
 			) }
 
 			{ ! isExample && (
-				<EditorNotice isDismissible={ false } icon="visibility">
+				<EditorNotice
+					isDismissible={ false }
+					icon="visibility"
+					componentActions={ [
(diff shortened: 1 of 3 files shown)


digitalwaveride (approved):
Works for me on our current live blocks:
image


Can't see it yet on the feedback block... but I assumed that it should not yet (... cause it is on the "feedback-block" branch, I was not sure if it should work there already)
image

PR #78, opened 2021-04-07
Add free signal notices

CGastrell:
This PR adds the notices we use for exhausted signals and errors while trying to sync to the Crowdsignal platform

## Test instructions
Checkout the branch and run `make client`

Use a free account with signals exhausted.

Insert a Feedback block. You should see the signal warning when you open the popup.

Disconnect, then make a change on the Feedback block settings. After 3 attempts (around 10-15 seconds) the sync error should appear on the popup. Connect again and hit the "Retry" button. Confirm your changes successfully synced by checking your dashboard.

### DIFF
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -12,6 +12,7 @@ import { RichText } from '@wordpress/block-editor';
 import { TextControl, TextareaControl } from '@wordpress/components';
 import { compose } from '@wordpress/compose';
 import { withSelect, dispatch } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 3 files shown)


digitalwaveride (changes requested):
Very nice Christian!
The API connection etc. seem to work for "out of signals" and "no connection" 👍 

Two little things I noticed:

- There is a little glitch in the styling of the notification, the text colors or in parts "white":

image

image

image

- If possible, the notifications should be outside of the card, the "popup UI content". The notifications are part of the editor UI like the toolbar, and should be below the toolbar and above the card (the red line), similar how we handle it for the NPS block:

image

digitalwaveride:
Thank you Christian!!! Awesome that things worked out with the notification banner placement! Nice! 😍



On more round of fine-tuning, but we are close:

### Colors 

image

### Icon Size

Could we give the icons the same size (the bigger one), could we use the same icon type? Whatever the "standard" Gutenberg one would be.
image

### Spacing & Alignment

Not sure if you have here some impact on the toolbar placement, but it would be super nice, if we could adjust this a bit:

image


### Three row signals notification banner -> any chance to bring it into two?

image

CGastrell:
All things noted, but these are the default notices we're using. Shall we need to address design issues on them, I'd like to do it on a separate card/PR. I do agree these things need to be fixed, but let keep things moving.

I created c/KtcbhyoS-tr/1284-improve-our-notice-warnings-design to keep track of this

digitalwaveride (approved):

PR #79, opened 2021-04-08
Add a backdrop behind the feedback block when selected inside the editor

ice9js:
This patch adds a backdrop to isolate the block from the background whenever it's selected inside the editor.  
It also addresses the issue with content popping in front of the block.

Before:

Screen Shot 2021-04-08 at 1 56 16 PM

After:

Screen Shot 2021-04-08 at 1 40 22 PM

# Testing

Verify the backdrop display correctly whenever the block is selected and clicking on it dismisses the block.

### DIFF
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -201,7 +201,7 @@ const EditFeedbackBlock = ( props ) => {
 			>
 				{ isSelected && (
 					<>
-						// eslint-disable-next-line
+						{ /* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions */ }
 						<div
 							aria-modal="true"
 							role="dialog"
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -165,6 +165,9 @@ const EditFeedbackBlock = ( props ) => {
 
 	const setPosition = ( x, y ) => setAttributes( { x, y } );
 
+	const toggleBlock = () =>
+		dispatch( 'core/block-editor' ).clearSelectedBlock();
+
 	const handleChangeAttribute = ( key ) => ( value ) =>
 		setAttributes( { [ key ]: value } );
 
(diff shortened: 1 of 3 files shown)


client/blocks/feedback/edit.js
```diff
@@ -147,45 +150,54 @@ const EditFeedbackBlock = ( props ) => {
...
-
-						<TextareaControl
-							className="crowdsignal-forms-feedback__input"
-							rows={ 6 }
-							onChange={ handleChangeAttribute(
-								'feedbackPlaceholder'
-							) }
-							value={ attributes.feedbackPlaceholder }
+					<>
+						// eslint-disable-next-line
```
CGastrell, line 154: Replace with:

`{ /* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions */ }`

client/blocks/feedback/edit.scss
```diff
@@ -4,6 +4,22 @@
 
 	.wp-block[data-type="crowdsignal-forms/feedback"] {
 		width: auto;
+		// This number is was carefully chosen so the block and its overlay
```
CGastrell, line 7: Present past contiguous tense is not allowed!

CGastrell (changes requested):
Only requesting changes because of the `//eslint-disable...` showing on the editor UI.

@digitalwaveride the "can't write on the feedback input (textarea)" issue is solved once this branch is rebased, it was fixed on #76 

Regarding the usual annoyances with the block/editor outlines, we can either ignore (as it might be out of our reach) or address on next iterations.

One thing that came to mind (out of the scope of this PR) is the "up & down" arrows on the toolbar, they don't make much sense on our case. They "work" (it doesn't break), but have no use. Are those removable? (I think not, but worth asking).

client/blocks/feedback/edit.scss
ice9js: Don't give them any more ideas :joy:

ice9js:
>One thing that came to mind (out of the scope of this PR) is the "up & down" arrows on the toolbar, they don't make much sense on our case. They "work" (it doesn't break), but have no use. Are those removable? (I think not, but worth asking).

Yeah we were thinking about this with @digitalwaveride but unfortunately there's no way to control these right now - Gutenberg just always adds them to the toolbar.

CGastrell (approved):
Nice work Kuba! LGTM!

PR #80, opened 2021-04-12
Update feedback block icons

ice9js:
This patch updates the feedback block with fresh icon for the block, trigger and the placement icons.  
For the placement icons, the icon in the toolbar should change according to the current block placement. I also removed the vertically centered options since they aren't quite so attractive with our current button option - as previously discussed with @digitalwaveride.

Screen Shot 2021-04-12 at 9 02 06 PM

The signal button icon is larger at 75px which affects the placement relative to the popover but I will be handling this separately as I'm redoing the whole positioning for the client side anyway.

# Testing

Verify the new icons are loading and show up as expected.


CGastrell (approved):
Looking good! :)

PR #81, opened 2021-04-12
Add a submit view for the feedback block

ice9js:
This patch adds a submit view to the feedback block. For now it just contains an editable thank you message.

Screen Shot 2021-04-12 at 10 13 57 PM

# Testing

You should now be able to toggle the feedback block between question and submit view inside the editor. The thank you message is editable.  
When you save and preview the post, after clicking submit on the feedback form the thank you message should be displayed instead of closing the popup.


CGastrell (approved):
LGTM!

PR #82, opened 2021-04-12
Change feedback default title

CGastrell:
This PR will change how the Feedback block title is set and defaulted.

The changes add a default title "Untitled Feedback" among the attributes. As soon as you add a Feedback block, the default title can be seen on the sidebar.

This change will make `title` and `header` (the prompt shown on the feedback modal) independent from each other. If the Title is set to empty, then it will default to the prompt as it we don't (can't) allow for empty titles, as opposed to default to the feedback `placeholder`.

## Test instructions
Checkout and `make client`. Insert a new Feedback block.

  - The title shown on the sidebar should be "Untitled Feedback".
  - The title should be the same that is shown on the Crowdsignal dashboard
  - Using an empty title should update the survey title with the `header` attribute (the modal prompt)

### DIFF
--- a/client/blocks/feedback/attributes.js
+++ b/client/blocks/feedback/attributes.js
@@ -51,7 +51,7 @@ export default {
 	},
 	title: {
 		type: 'string',
-		default: __( 'Untitled Feedback', 'crowdsignal-forms' ),
+		default: '',
 	},
 	x: {
 		type: 'string',
--- a/client/blocks/feedback/attributes.js
+++ b/client/blocks/feedback/attributes.js
@@ -51,7 +51,7 @@ export default {
 	},
 	title: {
 		type: 'string',
-		default: '',
+		default: __( 'Untitled Feedback', 'crowdsignal-forms' ),
 	},
 	x: {
 		type: 'string',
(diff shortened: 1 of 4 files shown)


client/blocks/feedback/attributes.js
```diff
@@ -51,7 +51,7 @@ export default {
 	},
 	title: {
 		type: 'string',
-		default: '',
+		default: __( 'Untitled Feedback', 'crowdsignal-forms' ),
```
ice9js, line 54: If we give it a default value the condition below where it takes either the title or the header value if no title is set would never work. We probably don't want this.

Also remember to make sure any changes are reflected in the PHP version.

client/blocks/feedback/edit.js
```diff
@@ -126,7 +127,7 @@ const EditFeedbackBlock = ( props ) => {
 					emailPlaceholder: data.emailPlaceholder,
 					sourceLink: data.sourceLink,
 					surveyId: data.surveyId,
-					title: data.title || data.feedbackPlaceholder,
+					title: data.title || data.header,
```
ice9js, line 130: 👍

ice9js (changes requested):

client/blocks/feedback/attributes.js
CGastrell: Agreed. There was some back and forth about it with Thomas, but I think this will work better. Easy change otherwise

ice9js (approved):
🚢  it!

PR #84, opened 2021-04-13
Refactor feedback block positioning

ice9js:
This patch is a major refactor of the positioning logic for the feedback block.  

I scrapped our previous css based solution and replaced it with some javascript. The `getFeedbackButtonPosition` also accepts a bounding DOM node as the last argument, making it reusable between the client and the editor which results in a much more consistent feel.

For now, I went with WordPress' native `<Popover />` component to power the popover on the client side as it fits the current design nicely. For alternative button/popover styles we'll probably want to implement our own but no need to do this right now.

Doing everything in JS also allowed me to implement a hopefully better solution for when the block is top-aligned in the editor where it previously had to be offset to the bottom to prevent the toolbar from getting obstructed by the header. Right now, the trigger will be displayed in the correct position and will only move down a bit when the block is selected. It's actually not nearly as distractive as I would've thought previously.

Bonus point: the popover will now close when you click outside of it on the client, however the button is acting a little weird which seems to be a bug with the actual `<Popover />` component from core. I'll need to find another way to address it.

Screen Shot 2021-04-13 at 10 08 59 AM

# Testing

Create a post with a feedback block. Try various position options and verify the block feels consistent across them in both the editor as well as on the post page.

### DIFF
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import React, { useLayoutEffect, useEffect, useState } from 'react';
+import React, { useLayoutEffect, useEffect, useRef, useState } from 'react';
 import classnames from 'classnames';
 import { get } from 'lodash';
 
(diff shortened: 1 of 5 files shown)


CGastrell (approved):
Feels more consistent, the bonus side effects also feel as a better UX :)

PR #85, opened 2021-04-13
Revamp notices

CGastrell:
This PR makes our notices more consistent and aligned.

`RetryNotice` component was created for less fuzz and the custom icon from the `SignalWarning` was removed in favor of the *built in* icon behavior of Notices through `icon` prop.

Note: we talked about removing the carriage return from the `SignalWarning` message (to prevent a 3 line message on the Feedback block), but it just looked too odd. Maybe it would be a good option to turn the second link into a button?
Suggested change **(not actually included on this PR)**:
image

## Test instructions
Insert a Feedback block and an NPS block. Force notices to show up by:

  - use a free account with exhausted signals
  - either block requests from your sandboxed API by `sudo global-http disable` or go *offline* from the Developer Tools, then make a change on the blocks (title?) to trigger a sync request (10-15 seconds later the new `RetryNotice` should appear)
  - on the NPS block, the threshold notice should always be present and in line with the other 2 notices (when shown)

Preview:
image

### DIFF
--- a/client/components/signal-warning/index.js
+++ b/client/components/signal-warning/index.js
@@ -11,7 +11,19 @@ import EditorNotice from 'components/editor-notice';
 
 const SignalWarning = () => {
 	return (
-		<EditorNotice icon="warning" status="warn" isDismissible={ false }>
+		<EditorNotice
+			icon="warning"
+			status="warn"
+			isDismissible={ false }
+			actions={ [
+				{
+					label: __( 'Please upgrade', 'crowdsignal-forms' ),
+					url: 'https://crowdsignal.com/pricing',
+					className: 'is-secondary',
+					noDefaultClasses: true,
+				},
+			] }
+		>
 			{ __(
 				'Your free Crowdsignal account has exceeded ',
 				'crowdsignal-forms'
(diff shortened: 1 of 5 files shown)


client/components/signal-warning/index.js
```diff
@@ -8,14 +8,21 @@ import { __ } from '@wordpress/i18n';
  * Internal dependencies
  */
 import EditorNotice from 'components/editor-notice';
-import WarningIcon from 'components/icon/warning-circle';
 
 const SignalWarning = () => {
```
ice9js, line 12: Nit: maybe we should call it `SignalNotice`/`SignalCountNotice`/`SignalWarningNotice` whichever you prefer... to keep it consistent with other `Notice` components?

We could also potentially put this and `retry-notice` in subdirectories inside `components/notice` since they're basically just specializations of that component. But maybe let's hold off on that for the repo cleanup.

ice9js (approved):
Works for me. Just one remark but overall good to go.

client/components/signal-warning/index.js
CGastrell: Yeah, crossed my mind as well. These are all notices and should be better named/arranged :)

PR #86, opened 2021-04-15
Add email feedback switch

CGastrell:
This PR adds the toggle to enable email notifications on every feedback submit, enabled by default.

## Test instructions
Checkout and `make client`. Test on docker against a sandboxed API D60245-code
Insert a Feedback block. Notice the toggle on the sidebar. Publish the post, visit and send feedback.
You should get an email with the feedback.
Edit the post, disable the toggle and send another feedback. This time you shouldn't get any email about it.

### DIFF
--- a/client/blocks/feedback/attributes.js
+++ b/client/blocks/feedback/attributes.js
@@ -90,4 +90,8 @@ export default {
 		type: 'string',
 		default: null,
 	},
+	emailResponses: {
+		type: 'boolean',
+		default: true,
+	},
 };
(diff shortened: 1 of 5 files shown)


ice9js (approved):
Works for me. Thanks @CGastrell!

digitalwaveride (commented):
Thanks Christian! Works 💘 !

One question: Would it be hard to let the user know to which address the email is sent to?

Showing the email something like this: 
image

Even if the email is long as in the example above, and not fitting into the line (I would suggest not to add a third line, no line break), the user would still get the idea where the email gets sent to.

PR #88, opened 2021-04-16
Enable custom feedback trigger images

ice9js:
This PR adds support for uploading custom images for the feedback dialog trigger button.  
There's now a `Feedback Button` section in the sidebar which will allow you to select a custom background for the dialog trigger button from your WordPress media library.

Screen Shot 2021-04-16 at 8 20 00 PM

# Testing

Verify the new section in the sidebar is displayed correctly.
Try uploading an image and verify it's displayed on the post as well.
Removing the image should revert to the default 'signal button'.

### DIFF
--- a/client/blocks/feedback/attributes.js
+++ b/client/blocks/feedback/attributes.js
@@ -53,6 +53,18 @@ export default {
 	textColor: {
 		type: 'string',
 	},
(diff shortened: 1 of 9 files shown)


CGastrell (approved):
Nice!!! Works beautifully! Thanks Kuba!

PR #90, opened 2021-04-20
Fix submit button hover

CGastrell:
The submit button inherits a `:hover` behavior that turns it invisible on 2021 theme. This PR applies the button color to the button wrapper text so the underlying button inherits a "viewable" color.

Try to read the above again, or watch the demos below.

Before:
hover-before

After:
hover-after

The change doesn't seem to affect other themes, like 2020, but worth checking.

## Test instructions
Checkout and `make client`. Visit a post with a Feedback block while using 2021 theme. Confirm the "Submit" button is visible when hover'd.

### DIFF
--- a/client/components/feedback/style.scss
+++ b/client/components/feedback/style.scss
@@ -56,6 +56,10 @@
 	justify-content: flex-end;
 	margin-top: 35px;
 	width: 100%;
+
+	&:hover {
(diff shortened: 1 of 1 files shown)


digitalwaveride (approved):
Nice find Christian, and even nicer fix! ;) 🚀

PR #92, opened 2021-04-22
Feedback block UX touch-ups

ice9js:
This patch adds a bunch of little UX touch-ups for the feedback block.  
Unfortunately a lot of them are somewhat related - stemming from the same core issue or building on top of each other. Hence the large PR.

I believe this solves c/IFakQEas for the most part.

To make things more manageable and easy to read I split up the frontend into multiple files. I'd like to do the same to the editor part at some point as well but that's not part of this change set.

Of the things that have been addressed, it's mostly popover/toggle interactions and some minor styling changes/fixes:

- Fixed issues with closing popovers by using `<Dropdown />` instead `<Popover />`
- The position popover will now close automatically after selecting a new position so it doesn't cover up the toggle button.
- I updated the position popover buttons to more closely resemble `<AlignmentMatrixControl />` from core. The icon is still to be updated.
- Fixed an issue where closing the block in the editor on Firefox would still maintain focus on the toggle making it impossible to re-open it without clicking somewhere else first.
- Removed block outlines in the editor.
- Removed the popover outline in the frontend as it didn't align with the actual block outline.
- Updated placeholder text for the inputs.
- Updated placeholder styling in the editor.
- Fixed input text size styles in the editor and the frontend.
- Fixed cursor style on the submit button in the editor.
- The feedback popover will now maintain the same hight for both feedback and submit views.
- Added missing block description.
- Added a checkbox in the Feedback Button section of the sidebar to trigger the shadow on the button on and off.
- Fixed the positioning of the toggle on some themes causing it to get overlapped by the admin header bar.
- The toggle will switch to a `close` button when the popover is active - frontend only.
- Fixed button text color setting not working correctly.
- Closing the popover will now reset the form. Regardless if it has been submitted or not.

# Testing

Verify the block builds correctly and that the above issues have been addressed.


digitalwaveride:
Yay! So many fixes! Thanks so much Kuba !!! 🚀 

Two things I noticed so far:


- There is a glitch with the card height for me... submit button outside of the card... 

image


-
> The toggle will switch to a close button when the popover is active - frontend only.

Not seeing this here yet... 

image

digitalwaveride:
The input fields could use more padding... is that an inhertied "global" theme class for the input field?

`.components-textarea-control__input `

Something like that would be great:

` padding:16px 16px;`

CGastrell (approved):
Nice work, lots of fixes and improvements!

The trigger button seems to get shifted when opening/closing the popup:
trigger-shift

Also, the "close" link could maybe not have the underline decoration on hover:
trigger-hover


Other than that (completely addressable on a different PR) LGTM!

digitalwaveride:
😍Nice Kuba!!! Thanks so much! 

Just one little thing, in the editor the padding below the Submit button is a bit big... it should be just 24px like in the published view.

image

PR #94, opened 2021-04-23
Update placement toolbar icon based on block position

ice9js:
This patch updates the icon of the placement popover to always reflect the current block position.

Screen Shot 2021-04-23 at 8 31 17 PM

# Testing

Try changing the block position and verify the correct dot is 'selected' on the icon.


digitalwaveride (approved):
Awesome Kuba! Very nice!
Good to go!

Saw by accident this glitch... the banner showing up below the block... but I guess this Is not related to this PR:

image

PR #96, opened 2021-04-23
Fix notices placement in the editor for the feedback block

ice9js:
This patch fixes an issue with notices appearing below the block when it's positioned on the bottom.

Before:

image

After:

Screen Shot 2021-04-23 at 9 03 55 PM

# Testing

Trigger a notice by, for example, going into offline mode and making changes in the block. Make sure it appears correctly.


CGastrell (approved):
LGTM!

digitalwaveride (approved):
Thanks Kuba for the quick fix!
Works! 🙏

PR #97, opened 2021-04-26
Fix close button hover state when a custom image is used

ice9js:
This change fixes the issue with the close button for the feedback block where it'd not change background correctly when activated while a background image is set.

# Testing

Create a feedback block and set a custom image for the feedback button.  
Verify that when the block is triggered, the button displays an 'x' icon on a solid background - same color as the other buttons.

### DIFF
--- a/client/components/feedback/style.scss
+++ b/client/components/feedback/style.scss
@@ -28,7 +28,7 @@
 	}
 
 	&.is-active {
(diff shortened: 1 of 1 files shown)


CGastrell (approved):
LGTM! There's still an issue on 2021 theme where the X becomes the same color as background on hover, but that's a different thing.

:shipit:

PR #98, opened 2021-04-26
Remove icons SVG fills

CGastrell:
Removing the `fill` property allows for CSS to rule over it, making those icons look more *aligned* with the rest out there.

Note that the Feedback Button icon was changed to use `fill` instead of `stroke`

**Before:**
fills-before

**After:**
fills-after

## Test instructions
Checkout and `make client`. Edit a post, type `/crowd` to get all of our blocks on the popup list. Hover the blocks in the list, the icons should get the same color of the text.

Check that Applause and Vote icons preserve functionality and styling on the frontend (Applause allows for a lot of color customization)


digitalwaveride (approved):
🚀

client/components/icon/applause.js
```diff
@@ -17,7 +17,7 @@ export default () => (
...
-				fill="black"
+				fill=""
```
ice9js, line 20: Why not just omit the attribute entirely?

ice9js (changes requested):
I'd say let's just remove the `fill` attributes although it looks like you just replaced some of the icons entirely?

In general we should maybe reconsider how we generate these... as in how we can actual export SVGs optimized for web use. The default exports are a mess.

ice9js (approved):
👍

PR #100, opened 2021-04-27
Use standard button for popup trigger

CGastrell:
This PR

  - removes the media upload section for the trigger button and related code
  - makes the trigger back into a button
  - apply styles to trigger button

Preview:
new-trigger

## Test instructions
Checkout and `make client`. Insert a Feedback block on a post.

On the Feedback block sidebar "Feedback button" section, there should be no media upload option, only the shadow toggle.

The trigger should be an editable button, customize the text and confirm on public frontend

~~Note: so far the button is a standard `isPrimary` button, this will probable change when we add the "edit button styling" feature~~

NOTE 2: a *cascade* of styles is applied to our buttons, we might want to improve this in the future. Right now the button is wrapped with `wp-block-button`, inherits from `wp-block-button__link` and is customized with `crowdsignal-forms-feedback__trigger` and `crowdsignal-forms-feedback__feedback-button`


digitalwaveride:
Thanks Christian for going after this! 🙏 


- Editor View: When the card is open, the cursor above the button should be "text" so users get a hint that the text is editable. (when the card is closed, it should be "pointer" of course)

- The styling of the buttons (text styling) does currently differ in the editor and published view
image

- Published View: The "close" button "x" is not yet properly implemented - currently there Is just a color change on the button bg, which would make no sense in that context. 
image
We should keep the color of the button and of course the size.
In order to make the button less competitive to the "Submit" button that might be just a few centimeters above it, we could experiment to give it a: 
opacity: 0.7;

- The Styling of the feedback button
As discussed in our call, for the button styling we would need to work with a mix of "dynamic" style elements inherited by the theme, and a few fixed ones.

FIXED:
-> font-size: 16px;
-> line-height: 150%;
-> text-align: center;
-> padding: 12px 16px;

INHERITED FROM THE THEME:
from the "button block" styling:
-> font-color
-> color
-> font-family
-> font-weight
-> border-radius
-> text-transform
-> text-decoration


- Published View: Noticed that if you use emojis in the text, in the published view the empty space after an emoji is ignored  
image

digitalwaveride:
Regarding the design of the Close state button ... it could look like this, the label "Close" with a X icon: 

image

client/blocks/feedback/edit.js
```diff
@@ -9,21 +9,20 @@ import { get } from 'lodash';
...
  * Internal dependencies
  */
 import ConnectToCrowdsignal from 'components/connect-to-crowdsignal';
-import SignalIcon from 'components/icon/signal';
 import { withFallbackStyles } from 'components/with-fallback-styles';
 import { getFeedbackButtonPosition } from 'components/feedback/util';
 import { useAccountInfo } from 'data/hooks';
 import Sidebar from './sidebar';
 import Toolbar from './toolbar';
-import { getStyleVars, getTriggerStyles } from './util';
```
ice9js, line 26: If `getTriggerStyles` isn't used anywhere anymore (which I think it shouldn't be at this point) let's remove it.

ice9js (changes requested):
I noticed we can remove a few more things and I'm not sure I understand how using Gutenberg's `<Button>` for the toggle is helpful for our case?

ice9js (approved):
>NOTE 2: a cascade of styles is applied to our buttons, we might want to improve this in the future. Right now the button is wrapped with wp-block-button, inherits from wp-block-button__link and is customized with crowdsignal-forms-feedback__trigger and crowdsignal-forms-feedback__feedback-button

Yeah, we'll probably want to grab these styles together with everything else eventually or we end up with the same issue. Good enough for now though.

Just one linter error remaining.

digitalwaveride (changes requested):
Thanks Christian for all the fixes! Nice! Looks much better!

We are close! Two things for fine-tuning:

- The Close button state, could we add there the "X" icon instead of the font letter?
image

- The Button Styling
It seems that it currently mirrors completely the "Submit Button", only the hover state is for me missing, but does not apply yet the fixed "hard-coded" css elements. 
image

image

digitalwaveride (approved):
👍

PR #101, opened 2021-04-27
Change sidebar URL for Crowdsignal

CGastrell:
This PR

  - changes the URL to crowdsignal.com so it doesn't mutate to that of the results.

## Test instructions
Checkout and `make client`. Insert all our blocks on a post. The link to Crowdsignal on the "Results" section of the sidebar should always point to crowdsignal.com


digitalwaveride (approved):
Nice! Thanks so much Christian! 🚀

PR #102, opened 2021-04-29
Fix feedback block's editor overlay issues on Safari

ice9js:
Addresses c/r5SXo3AO-tr. The issue only appears on Safari where the overlay covers portions of the editor UI such as the header bar or the sidebar when the feedback block is active due to Safari's `z-index` implementation.

Since we don't have control over the rest of the editor, I updated our block to manually position the overlay and only cover the content part of the editor.  
While at it I also fixed another issue I noticed, which was the button appearing too low if full screen mode was off.

## Testing

Verify the block overlay no longer covers editor controls on Safari.  
The button should now be correctly positioned in the editor when using a bottom position and with full screen mode off.

### DIFF
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -160,7 +160,7 @@ const EditFeedbackBlock = ( props ) => {
 			right: window.innerWidth - ( contentBox.left + contentBox.width ),
 			top: contentBox.top,
 		} );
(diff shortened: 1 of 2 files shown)


digitalwaveride (approved):
Looks nice! Works 🚀🚀🚀 
Thanks Kuba!

image

PR #103, opened 2021-04-29
Prevent inputs from overflowing the popover on legacy themes

ice9js:
Fixes c/LsSBBKkQ-tr. Looks like some of the old themes have some old-school styling rules as one might expect which is causing the input and the textarea to overflow the dialog.
Forcing `box-sizing: border-box` manually fixes the issue.

# Testing

The issue was found on Twenty Eleven. Install and activate the theme and verify the input and the textarea now fit correctly.

### DIFF
--- a/client/components/feedback/style.scss
+++ b/client/components/feedback/style.scss
@@ -74,6 +74,7 @@
 
 	input,
 	textarea {
+		box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


digitalwaveride (approved):
Thanks Kuba for the fix! Works!!! 🥳

PR #104, opened 2021-04-29
Fix title input for feedback and nps blocks

ice9js:
This issue was spotted with the feedback block but the fix addresses it for both feedback and NPS block which seems to have a similar issue.  
With this fix, the title input will initially contain the feedback question text until it's edited. After editing it, it'll always be the value that it was set to - including empty. Previously it'd revert back to the question text when you try deleting the value.

# Testing

Verify the feedback and NPS title inputs now work as expected.

### DIFF
--- a/client/blocks/feedback/attributes.js
+++ b/client/blocks/feedback/attributes.js
@@ -76,7 +76,7 @@ export default {
 	},
 	title: {
 		type: 'string',
-		default: null,
+		default: '',
 	},
 	x: {
 		type: 'string',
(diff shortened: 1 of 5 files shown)


digitalwaveride:
In this iteration, the field for the title input field is by default filled with the text of the question. 

I am wondering if this brings more confusion then it might help. The reason is that it could set the expectation that this is the place where you also can change the "question of the block" ... a few users might make some conclusions here that are not intended.

If the input field is empty by default though, till it is actually overwritten, then users will not 

For the cases where the title field stays empty, we would show, as before, the text of the question as title in the dashboard.

digitalwaveride (approved):
Works like a charm! Thanks Kuba! 🚀

PR #106, opened 2021-04-29
Change feedback block description

CGastrell:
This PR updates the block description and adds an example to be rendered on the block library search preview

## Test instructions
Checkout and `make client`. Use the block library search to find the Feedback block, hovering it should show a preview with default values.

Known issue: the backdrop renders inside the preview somewhat quirky, but all this sizes and positioning are being skipped to be able to render the preview.


digitalwaveride (approved):
Thanks Christian! 🚀 

The only thing that I noticed is that it is not showing up reliably...
and can reproduce a weird behaviour, not sure though if that is in our hands or an issue on the Gutenberg side.

When my mouse comes from the top and moves on the block, I see in most cases this:

image

But when I move my mouse on the block coming from the bottom I see only this... 
🤷 

image

See Video:

https://user-images.githubusercontent.com/5120357/116687525-a22d8900-a9b5-11eb-91eb-5990efa1471d.mov

PR #107, opened 2021-04-29
Add close status notification

CGastrell:
This PR adds a notice at the bottom of the popover to indicate if the Feedback block is closed or its expiry date has passed

## Test instructions
Checkout and `make client`. On a Feedback block, change settings to "Closed" or "Closed after" and a past date. The notice should be noticeable

### DIFF
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -201,7 +201,7 @@ const EditFeedbackBlock = ( props ) => {
 		height,
 	};
 
-	const isCurrentlyClosed =
+	const isClosed =
 		FeedbackStatus.CLOSED === attributes.status ||
 		( FeedbackStatus.CLOSED_AFTER === attributes.status &&
 			null !== attributes.closedAfterDateTime &&
@@ -314,7 +314,7 @@ const EditFeedbackBlock = ( props ) => {
 								/>
 							</div>
 						) }
-						{ isCurrentlyClosed && (
+						{ isClosed && (
 							<div className="crowdsignal-forms-feedback__closed-notice">
 								{ __(
 									'This Feedback Form is Closed',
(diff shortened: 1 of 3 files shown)


digitalwaveride (changes requested):
Thanks Christian for bringing this in so fast!!! 🙏 

Works for me!
Just some styling fine-tuning would be great. Looking at this right now... 

image



- The color of that bar is a bit an issue as it is squeezed between the two buttons. We could use the same color as the buttons, but than it would maybe not stand out enough.
But better would be maybe to go with neutral standard Gutenberg "dark color" that should work with most Submit button colors etc: 
background-color: #1e1e1e

- For the text-styling, could we go with:

```
font-size: 14px;
line-height: 150%;
padding: 6px;
```

- The Text
I would keep it a tiny bit simpler, and mimic the same statement that we have on the poll block for that closed state:

`This Feedback Form is Closed`

All together could like this:

image

client/blocks/feedback/edit.js
```diff
@@ -200,6 +201,12 @@ const EditFeedbackBlock = ( props ) => {
 		height,
 	};
 
+	const isCurrentlyClosed =
```
ice9js, line 204: Nit: Not a huge fan of overly descriptive variable names, let's maybe cut it down to `isClosed`?

client/components/feedback/style.scss
```diff
@@ -4,6 +4,11 @@
 	display: flex;
 	position: fixed;
 	z-index: 100;
+
+	> div:last-child:not(.crowdsignal-forms-feedback__trigger) {
```
ice9js, line 8: What is this and why can't we just give it a proper class name?

ice9js (commented):
Overall 👍  I left two small comments.

digitalwaveride (approved):
😍  🚀

client/components/feedback/style.scss
CGastrell: We need the rounded corners at the bottom of the modal, and the new notice that goes there is just appended to the popover. So, whatever div is last, excluding the `trigger`, should have this roundness.

The notice does have a class `.crowdsignal-forms-feedback__closed-notice`, but if we just put the rounded corners on both it doesn't look good, like a double smile on a face.

PR #108, opened 2021-04-30
Feedback block editor UI fixes

ice9js:
This patch contains the fixes for a number of issues found with the feedback block's editor UI:

- The background on the feedback/submit view toggle should now always appear centered behind the text.
- The placement icon dots have been updated to be a little bigger to make them more clear.
- Fixed the issue with the block height changing when switching to the submit view right after selecting the block.
- Updated the feedback block icon.

# Testing

Verify the above issues have been addressed and the new icons look ok.


digitalwaveride (approved):
😍 Awesome! Thanks Kuba!

Two things that might be not part of this diff:

Not sure if these fixes should already include the "Tool Tip" for the Postioning icon in the toolbar - we should use the same text as for the cover block icon:

image

;) 


And one other thing I noticed... not necessarily part of this diff, but the space between the toolbar and the card feels really big now... did something change there?


See for example Cover Block:

image


Vs. Feedback Button block:

image


Any chance to adjust that, to some "standard Gutenberg" value?

🙏

ice9js:
I have added the tooltip but updated the text to `Change toggle button position`. Gutenberg's cover blocks `Change content position` makes little sense in this context.

As for the other margin/spacing issue, it seems it was introduced in #100. I'll add a fix for it separately.

digitalwaveride:
Thanks Kuba! 🙏 

How about making the text even simpler?

`Change button position`

Most users likely don't know what a toggle is anyways...

digitalwaveride (approved):
🚢

PR #112, opened 2021-04-30
Refresh readme

CGastrell:
Just a kickstarter to improve our Readme.md

The contents are slightly different than those on README.txt mostly because the txt file responds to .org mandatory format.

### DIFF
--- a/Development.md
+++ b/Development.md
@@ -1,0 +1,89 @@
+
+# Crowdsignal Forms
+
+Note: this file is intended for developers, the plugin readme
+is [README.txt](README.txt)/[README.md](README.md)
+
+## Using docker for local dev
+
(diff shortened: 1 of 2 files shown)


Development.md
```diff
@@ -0,0 +1,89 @@
+
```
ice9js, line 1: Nit:

Perhaps it'd make more sense to add a `docs/` directory and put this in there under `development-environment.md`. As it stands this file is a little bit of everything :(
Instead it'd be nice if we could get a proper `CONTRIBUTING.md` into the root folder which just has some general contributing guidelines like, where to look for more information, a notice about coding standards or anything else one might want to keep in mind when starting to work on the plugin.

digitalwaveride (approved):

PR #134, opened 2021-05-07
Make branding enabled by default on atomic sites

CGastrell:
This PR adds a default value to be passed to the filter `crowdsignal_forms_branding_enabled` based on defined constants. 

## Test instructions
This is mostly theoretical, as I still have to find a way to debug Atomic sites. 

Ref: p1620403053188500-slack-C7YPW6K40


ice9js (approved):
👍

PR #135, opened 2021-05-07
Fix poll styling issues

CGastrell:
This PR

  - adds a line-height directive for the poll question, consistently rendering the same height on both editor and public frontend
  - remove redundant wrapper class on answers so editor and public frontend for a more wysiwyg experience

It will also address a long overdue fix: answers could be clicked anywhere on the row, but only recognize the click when clicking ON the label/input.

## Test instructions
Checkout and `make client`. Insert a Poll block. Activate 2019 theme. Compare editor and public frontend. Line height should show the same height on both.
Check the height/separation on the answers, they should display (almost) exactly the same height on editor and public frontend

### DIFF
--- a/client/components/poll/answer.js
+++ b/client/components/poll/answer.js
@@ -87,10 +87,10 @@ const PollAnswer = ( {
 	);
 
 	return (
-		<>
+		<div className={ classes }>
 			{ AnswerStyle.RADIO === answerStyle && renderRadioAnswers() }
 			{ AnswerStyle.BUTTON === answerStyle && renderButtonAnswers() }
-		</>
+		</div>
 	);
 };
 
--- a/client/components/poll/answer.js
+++ b/client/components/poll/answer.js
@@ -87,10 +87,10 @@ const PollAnswer = ( {
 	);
 
 	return (
-		<div className={ classes }>
+		<>
 			{ AnswerStyle.RADIO === answerStyle && renderRadioAnswers() }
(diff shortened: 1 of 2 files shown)


digitalwaveride:
That's great, thanks Christian!
Seeing the same spacings on both view now 👍 !

We need to finetune the spacings though. The hierarchy is the wrong way.
That's what I am seeing currently:

image

The distance between the Question and the Answer Group, needs to be bigger, than the distance inside the Answer Group's answer options.

In the scribble below, B needs to be smaller than A, so that we give the eye some guidance what belongs to each other.

image


Design file example:

image


Played with the CSS:

image


```
.crowdsignal-forms-poll h3.crowdsignal-forms-poll__question {
    margin: 0 0 44px;
}

.crowdsignal-forms-poll__answer:not(.is-button) {
    margin-block-end: 12px;
}

.crowdsignal-forms-poll__actions {
    display: flex;
    justify-content: flex-end;
    margin-top: 32px;
    margin-bottom: 16px;
}
```

Not sure if that would make sense in the code... but you get the idea, the distance between the answer options need to be smaller than the distances between the question and first answer, as well as the last answer and submit button.

digitalwaveride:
Thanks Christian! Looks good on the List view 👍 🙏 , unfortunately, breaks on the button view... 

image

digitalwaveride (approved):
Yay! Perfect!!! ❤️ 

Thanks Christian!

PR #137, opened 2021-05-10
Fix verify info request logic

CGastrell:
# Improves backend interaction with less requests

NOTE: enqueued for next release

This PR addresses a couple of hiccups when performing basic checks on the editor.

Mostly, this new implementation aims to be able to differentiate between a wrongly set up API Key and an unverified user. Right now, users with a wrong API Key would receive a notice stating the user has not verified its email, which leads to confusion among users and HEs.

One of the small refactors has been made on the *API gateway* file, which now uses the API request `/account/info` to get all related information and cache it for a minute. Related methods to get capabilities and user email verification status now rely on `get_account_info` and get the responses from there. This brings down the connection request checks from 2-5 seconds to 1s (uncached) and some 100-200ms once cached (testing on local, I expect to see improvements on remote scenarios as well).

All the logic made by ConnectToCrowdsignal component has been refactored to not use the particular request `/connected` and now evaluates the response from `/info` to derive the status.

## Test instructions
There are several use cases/scenarios we can test:

  - Failed connection to CS API: while working on docker/local environment, merely disable global http connections to CS API (sandboxed). This is unrecoverable and should never happen, but we want to react nicely just in case.
  - Setup the plugin to use a key, then delete the key: you should now get the right message stating that you should configure the plugin

### DIFF
--- a/client/blocks/applause/edit.js
+++ b/client/blocks/applause/edit.js
@@ -36,7 +36,7 @@ const EditApplauseBlock = ( props ) => {
 		setAttributes
 	);
 
-	const accountInfo = useAccountInfo();
+	const accountInfo = get( useAccountInfo(), 'data', {} );
 
 	const shouldPromote = get( accountInfo, [
 		'signalCount',
--- a/client/blocks/applause/edit.js
+++ b/client/blocks/applause/edit.js
@@ -36,7 +36,7 @@ const EditApplauseBlock = ( props ) => {
 		setAttributes
 	);
 
-	const accountInfo = get( useAccountInfo(), 'data', {} );
+	const { accountInfo } = useAccountInfo();
 
 	const shouldPromote = get( accountInfo, [
 		'signalCount',
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -30,6 +30,7 @@ import SignalWarning from 'components/signal-warning';
 import { views, FeedbackStatus } from './constants';
 import RetryNotice from 'components/retry-notice';
 import FooterBranding from 'components/footer-branding';
+import FeedbackIcon from 'components/icon/feedback';
 
 const EditFeedbackBlock = ( props ) => {
 	const [ view, setView ] = useState( views.QUESTION );
@@ -253,7 +254,10 @@ const EditFeedbackBlock = ( props ) => {
 	);
 
 	return (
-		<ConnectToCrowdsignal>
+		<ConnectToCrowdsignal
+			blockName={ __( 'Feedback Button', 'crowdsignal-forms' ) }
+			blockIcon=<FeedbackIcon />
+		>
 			<Toolbar
 				currentView={ view }
 				onViewChange={ setView }
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -256,7 +256,7 @@ const EditFeedbackBlock = ( props ) => {
 	return (
 		<ConnectToCrowdsignal
 			blockName={ __( 'Feedback Button', 'crowdsignal-forms' ) }
-			blockIcon=<FeedbackIcon />
+			blockIcon={ <FeedbackIcon /> }
 		>
 			<Toolbar
 				currentView={ view }
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -65,8 +65,6 @@ const EditFeedbackBlock = ( props ) => {
 	const triggerButton = useRef( null );
 	const popover = useRef( null );
 
-	const accountInfo = useAccountInfo();
-
 	const { error: saveError, save: saveBlock } = useAutosave(
 		async ( data ) => {
 			dispatch( 'core/editor' ).lockPostSaving( clientId );
@@ -210,6 +208,8 @@ const EditFeedbackBlock = ( props ) => {
 	const handleChangeAttribute = ( key ) => ( value ) =>
 		setAttributes( { [ key ]: value } );
 
+	const accountInfo = get( useAccountInfo(), 'data', {} );
+
 	const shouldPromote = get( accountInfo, [
 		'signalCount',
 		'shouldDisplay',
@@ -248,7 +248,7 @@ const EditFeedbackBlock = ( props ) => {
 			null !== attributes.closedAfterDateTime &&
 			new Date().toISOString() > attributes.closedAfterDateTime );
 
-	const hideBranding = get( accountInfo, 'capabilities' ).includes(
+	const hideBranding = get( accountInfo, 'capabilities', [] ).includes(
 		'hide-branding'
 	);
 
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -208,7 +208,7 @@ const EditFeedbackBlock = ( props ) => {
 	const handleChangeAttribute = ( key ) => ( value ) =>
 		setAttributes( { [ key ]: value } );
 
-	const accountInfo = get( useAccountInfo(), 'data', {} );
+	const { accountInfo } = useAccountInfo();
 
 	const shouldPromote = get( accountInfo, [
 		'signalCount',
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -116,7 +116,7 @@ const EditFeedbackBlock = ( props ) => {
 	}, [ isSelected ] );
 
 	useLayoutEffect( () => {
-		if ( isExample ) {
+		if ( isExample || ! triggerButton.current ) {
 			return;
 		}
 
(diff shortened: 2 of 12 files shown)


client/blocks/applause/edit.js
```diff
@@ -36,7 +36,7 @@ const EditApplauseBlock = ( props ) => {
 		setAttributes
 	);
 
-	const accountInfo = useAccountInfo();
+	const accountInfo = get( useAccountInfo(), 'data', {} );
```
ice9js, line 39: Nit: ```const { data: accountInfo } = useAccountInfo();```

Though it's not an issue with a simple `get()` I think it's best to avoid adding any logic wrappers around hook as it can cause rendering issues.

client/blocks/feedback/edit.js
```diff
@@ -174,6 +172,8 @@ const EditFeedbackBlock = ( props ) => {
 	const handleChangeAttribute = ( key ) => ( value ) =>
 		setAttributes( { [ key ]: value } );
 
+	const accountInfo = get( useAccountInfo(), 'data', {} );
```
ice9js, line 175: See above.

client/blocks/nps/edit.js
```diff
@@ -109,9 +109,9 @@ const EditNpsBlock = ( props ) => {
 		'is-inactive': ! isExample && ! isSelected,
 	} );
 
-	const accountInfo = useAccountInfo();
+	const accountInfo = get( useAccountInfo(), 'data', {} );
```
ice9js, line 112: See above.

client/blocks/poll/edit.js
```diff
@@ -124,9 +124,9 @@ const PollBlock = ( props ) => {
 	const isHidden =
 		isClosed && ClosedPollState.HIDDEN === attributes.closedPollState;
 
-	const accountInfo = useAccountInfo();
+	const accountInfo = get( useAccountInfo(), 'data', {} );
```
ice9js, line 127: See above.

client/blocks/vote/edit.js
```diff
@@ -56,7 +56,7 @@ const EditVoteBlock = ( props ) => {
 
 	const voteItemStyleVars = getVoteStyleVars( attributes );
 
-	const accountInfo = useAccountInfo();
+	const accountInfo = get( useAccountInfo(), 'data', {} );
```
ice9js, line 59: See above.

client/data/hooks/index.js
```diff
@@ -132,10 +101,11 @@ export const useAccountInfo = () => {
...
 		const info = await requestAccountInfo();
 		setAccountInfo( info );
+		return info;
 	};
 
 	useEffect( () => {
 		getAccountInfo();
 	}, [] );
-	return accountInfo;
+	return { data: accountInfo, reloadAccountInfo: getAccountInfo };
```
ice9js, line 110: I suggest you just expose `accountInfo` as `accountInfo` and not `data` - much more easy to understand when looking at it from the other end and simplifies destructuring as well.

client/data/hooks/index.js
```diff
@@ -132,10 +101,11 @@ export const useAccountInfo = () => {
 	const getAccountInfo = async () => {
 		const info = await requestAccountInfo();
 		setAccountInfo( info );
+		return info;
```
ice9js, line 104: Technically not really necessary since when `accountInfo` is updated that automatically propagates down to wherever this hook is used. So you shouldn't need to rely on `reloadAccountInfo()` to return a value. The function name doesn't suggest it either.

includes/frontend/class-crowdsignal-forms-block.php
```diff
@@ -73,6 +73,9 @@ protected function should_hide_branding() {
 			return true;
 		}
 
+		// TODO code review note:
+		// now that the gateway is caching get_account_info, hence, get_capabilities,
+		// should we remove this block?
 		if ( get_transient( self::TRANSIENT_HIDE_BRANDING ) ) {
```
ice9js, line 79: Sounds good.

includes/gateways/class-api-gateway.php
```diff
@@ -416,22 +416,11 @@ private function create_or_update_poll( Poll $poll ) {
...
-			if ( isset( $response_data['error'] ) ) {
-				return new \WP_Error( $response_data['error'], $response_data );
-			}
-			return new \WP_Error( 'decode-failed' );
-		}
+		$response_data = $this->get_account_info();
 
-		return $response_data;
+		return isset( $response_data['capabilities'] )
+			? (bool) $response_data['capabilities']
```
ice9js, line 422: Is this boolean conversion correct?

includes/gateways/class-api-gateway.php
```diff
@@ -524,45 +513,54 @@ private function is_poll_response_valid( $response ) {
...
-		if ( null === $response_data || ! is_array( $response_data ) ) {
-			if ( isset( $response_data['error'] ) ) {
-				return new \WP_Error( $response_data['error'], $response_data );
-			}
-			return new \WP_Error( 'decode-failed' );
-		}
-
-		return $response_data['is_verified'];
+		return isset( $response_data['is_verified'] )
+			? (bool) $response_data['is_verified']
```
ice9js, line 519: How about `return (bool) $response_data['is_verified'] ?? false;` ?

ice9js (commented):
Nice going Christian! For now I left some comments and questions.  
Eventually it'd be great to actually cache `accountInfo` on the frontend as well - since it's not something that's dependent on the parent component/block and then we'd have to do the request only once.

client/data/hooks/index.js
CGastrell: Completely agree, but did you see how it's used? Providing both data and a reloading mechanism makes for this scenario where you'd need to "try to connect again using the reloader, if successful, use the newly returned data". Not sure it can work as you suggest (which would be the optimal/react way of doing things)

ice9js: I see now, yeah... It probably makes a lot more sense for the entire hook to be replaced by a state branch with an appropriate _control_ and _resolver_ - in `@wordpress/data` terms.

But until we implement that, this should be ok.

includes/gateways/class-api-gateway.php
CGastrell: NIET!

includes/gateways/class-api-gateway.php
CGastrell: I have a warning here due to coalescent operator not being available on PHP <5.6, not sure this is a thing, but maybe just keep it as it is for now.

ice9js: Oh I totally forgot about WP's minimum supported version. My bad, sorry.

client/blocks/feedback/edit.js
```diff
@@ -248,12 +249,15 @@ const EditFeedbackBlock = ( props ) => {
...
-	const hideBranding = get( accountInfo, 'capabilities' ).includes(
+	const hideBranding = get( accountInfo, 'capabilities', [] ).includes(
 		'hide-branding'
 	);
 
 	return (
-		<ConnectToCrowdsignal>
+		<ConnectToCrowdsignal
+			blockName={ __( 'Feedback Button', 'crowdsignal-forms' ) }
+			blockIcon=<FeedbackIcon />
```
ice9js, line 259: Missing braces.

client/blocks/feedback/edit.js
```diff
@@ -248,12 +249,15 @@ const EditFeedbackBlock = ( props ) => {
 			null !== attributes.closedAfterDateTime &&
 			new Date().toISOString() > attributes.closedAfterDateTime );
 
-	const hideBranding = get( accountInfo, 'capabilities' ).includes(
+	const hideBranding = get( accountInfo, 'capabilities', [] ).includes(
```
ice9js, line 252: Nit, I think this would be a little more concise:

```
const hideBranding = includes( accountInfo.capabilities || [], 'hide-branding' );
```

client/blocks/feedback/edit.js
CGastrell: Wrapping `<FeedbackIcon />`? Weird it didn't blow up, but I'll them if that's the case.

client/blocks/feedback/edit.js
CGastrell: Isn't one of `lodash.get`'s missions to ensure the same output even if `accountInfo` is `undefined` or `null`? What I'm trying to say is the use of `get` is so we don't have to check `accountInfo && accountInfo.capabilities`, we just rely on `get` to try that and provide the default if it fails. Syntax does look more concise though, just making sure to not throw safe mechanisms out the window in favor of neat syntax.

That said, yes, I try to make sure `accountInfo` always has some value, but better safe than uncontrolled errors :)

ice9js (approved):
LGTM! :shipit:

Issue #129, opened 2021-05-05
Feedback Button Block - Clicking "Done" from Mobile Browser closes the entire form


rosepajaroja:
Clicking "Done" closes the entire form instead of moving to the next field or minimizing the keyboard. 

iOS iPhone X
Tested both on Safari and Chrome

Issue #125, opened 2021-05-05
Feedback button: Email field doesn't have a validation


sanjeev00733:
It would be great to have a message asking users to enter the right email address if they are entering it in the wrong format.

Currently, I can enter anything in the email field and the form gets submitted, the recorded response automatically removes the email address field if it is not in the correct format but it would be good if we can notify the user in the first place to enter email in the right format.

bobmatyas:
**Verified — this appears to be resolved.** 👍

The email field on the Feedback block is now a native `<input type="email">` (`client/components/feedback/form.js`), which enables the browser's built-in format validation. Testing locally by submitting the feedback form with an invalid value like `notanemail`, the form is blocked from submitting and the browser prompts the user:

> Please include an '@' in the email address. 'notanemail' is missing an '@'.

This gives users the up-front notification requested here, rather than silently accepting and dropping the malformed address.

This was introduced in #263 (the field was previously a plain-text `TextControl` with no format validation).

One note: the validation is provided by the browser's native `type="email"` behavior, so its appearance is browser-styled rather than a custom in-form message. If a consistent, on-brand inline error is desired, that would be a good follow-up enhancement — but the original ask (prompt the user for a correctly formatted email) is now handled.

Recommend closing as resolved.

bobmatyas:
If the email isn't set as required and someone tries to submit, it runs the email check. Closing this as not reproducible.

Issue #99, opened 2021-04-27
Feedback Button: Provide an option to make email address mandatory


bsanevans:
Feature request: I was testing the Feedback Button. As a site owner, I wanted to make the email field mandatory - to prevent people spamming me, or anonymously ranting on my site. But this option does not exist right now.

Suggestion: Add a toggle in the block settings, to change whether the email field is mandatory or not.

zdenys:
I can see that the option to make email field mandatory is now present 🥳  

However, the placeholder text in the email field still appears as "Your email (optional)" on the frontend (not in the Editor).

https://user-images.githubusercontent.com/39308239/135997519-ce65cda6-83d4-4bc4-a04b-df8aab3b29ed.mov

| Editor view  | Frontend |
|-------|------|
| ![image](https://user-images.githubusercontent.com/39308239/135997753-247abf6e-653c-48fd-896e-9827311d1bce.png) | ![image](https://user-images.githubusercontent.com/39308239/135997884-6b7aa6cf-f208-42bf-a14a-93835878ce4e.png) |

**What I expect to happen is:**

- [ ] "Your email (optional)" both on frontend and backend only if `Require email address` option is toggled off. 
- [ ] "Your email" both on frontend and backend only if `Require email address` option is toggled off.

aicee:
Also reported in 5477162-zen, they are seeing "(optional)" text in the Email Address field placeholder even if `Require email address` option is enabled.

PR #138, opened 2021-05-11
Implement a toggle for making feedback email mandatory

ice9js:
This patch solves #99 by adding a toggle to make the email field in the feedback block required or optional under in the settings area of the sidebar. It also adds validation on the feedback block frontend, so if the feedback or the required email field are left empty, they should light up red after attempting to submit the form (#125).

I also used this opportunity to update the markup so it contains a proper `<form>` element which should help with possible submission issues that were reported in #129.

# Testing

Verify the new setting and validation for the email field works.

### DIFF
--- a/client/blocks/feedback/attributes.js
+++ b/client/blocks/feedback/attributes.js
@@ -28,6 +28,10 @@ export default {
 		type: 'string',
 		default: __( 'Your Email (optional)', 'crowdsignal-forms' ),
 	},
+	emailRequired: {
+		type: 'boolean',
+		default: false,
+	},
 	feedbackPlaceholder: {
 		type: 'string',
 		default: __(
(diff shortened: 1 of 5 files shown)


ice9js:
>I'd ask just a small adjustment: the input placeholder, both on editor and public view, still includes the text "(optional)" when toggling the option to make in mandatory.

I removed the `(optional)` label. Unfortunately toggling it on and off based on the setting is a little weird considering the text is user editable anyway. It could cause some confusion.  
But with the validation it should be less of a problem.

CGastrell (approved):
LGTM :shipit:

zdenys:
>I removed the (optional) label.

@ice9js I still see that and commented here with more details https://github.com/Automattic/crowdsignal-forms/issues/99#issuecomment-934238501

PR #139, opened 2021-05-11
Add different toggle settings for the feedback button block

ice9js:
This patch adds a new setting under the 'feedback button' section of the sidebar which allows to change how the feedback popup is triggered. The current options are:

- on click
- on hover
- on page load

Solves c/emTOHVUi-tr.

# Testing

Verify the new options work as described.

### DIFF
--- a/client/blocks/feedback/attributes.js
+++ b/client/blocks/feedback/attributes.js
@@ -6,7 +6,7 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
-import { FeedbackStatus } from './constants';
+import { FeedbackStatus, FeedbackToggleMode } from './constants';
 
 /**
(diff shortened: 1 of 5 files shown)


CGastrell (approved):
LGTM! :shipit:

digitalwaveride (changes requested):
❤️ 
Thanks Kuba! Works for me!

One adjustment regarding the placement in these sidebar: The Feedback Button area feels like a "styling section". How about putting it into the settings section, where there is also the "status" dropdown that defines also a basic setting. The loading behaviour setting fits there as well.

image

And one suggestion regarding the label naming, instead of "Toggle on" that sounds a bit technical and users might not know what a "toggle behaviour" is, how about:

`Show form on:`

image

digitalwaveride (approved):
Perfect! Works as advertised!
Thanks so much for the changes!
🚀 🚀

PR #140, opened 2021-05-12
Add a vertical middle position for the feedback block

ice9js:
This patch adds a vertically centered position for the feedback block.  
When this position is selected the button appears vertically and is positioned against the edge of the page and the popover should appear to the side of it.

As an additional improvement, I implemented our version of `<BlockAlignmentControl />` component which can be configured with any matrix size/configuration and can now be also navigated using the keyboard.

Screen Shot 2021-05-13 at 2 17 48 PM

# Testing

You'll need to run `npm install` after you apply the patch.
Test that the new positions work in both the editor as well as the published page.

### DIFF
--- a/assets/stylesheets/editor.scss
+++ b/assets/stylesheets/editor.scss
@@ -2,6 +2,7 @@
 @import "./shared/variables.scss";
 
 // Components
+@import "components/block-alignment-control/style.scss";
 @import "components/connect-to-crowdsignal/style.scss";
 @import "components/editor-notice/style.scss";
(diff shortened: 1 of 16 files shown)


digitalwaveride (changes requested):
Kuba, that looks very cool, and works perfectly... nearly... 

When you start editing the text of the button, things start to break unfortunately... 

image

digitalwaveride (approved):
CSS wizard of the week! 🎩 

🥳 
Thanks so much Kuba for hanging in there .... never giving up! 🚀

PR #142, opened 2021-05-13
Add filter to control transform behavior

CGastrell:
This PR is another PoC in the attempt to prevent Feedback block to be transformed into other blocks that don't make sense. In particular, this adds a filter preventing any transformation of the block. However, Jetpack's "Premium Content" still can't be prevented.

With the filter in place, Column and Group are no longer available as a "Transform to" option, which is a good start I guess. Feedback block just doesn't make sense inside a column (its layout won't respect the column restrictions) and Group might have some valid use case, but none so far demonstrating to be useful.

Shall we consider this a successful change, it should also be applied to NPS block. 

## Test instructions
Checkout and `make client`. Insert a Feedback block and try to use the leftmost toolbar button to transform it into some other block, it shouldn't even display a list.
On Jetpack enabled sites, the list will only show "Premium Content"

### DIFF
--- a/client/editor.js
+++ b/client/editor.js
@@ -35,3 +35,13 @@ addFilter(
 	'crowdsignal-forms/with-fixed-position-control',
 	withFixedPositionControl
 );
+addFilter(
+	'blocks.registerBlockType',
+	'crowdsignal-forms/feedback',
+	( settings ) => {
+		return {
+			...settings,
+			transforms: null,
+		};
+	}
+);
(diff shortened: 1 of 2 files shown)


client/editor.js
```diff
@@ -35,3 +35,13 @@ addFilter(
 	'crowdsignal-forms/with-fixed-position-control',
 	withFixedPositionControl
 );
+addFilter(
```
ice9js, line 38: Nit: Because this is directly related to the feedback block, maybe put it inside `client/blocks/feedback/index.js` ?

ice9js (approved):
Works for me. Just left a small comment about where it might make more sense to put the code.

PR #145, opened 2021-05-14
Make transform filter match the blockname

CGastrell:
The filter is affecting ALL blocks and needs to be more specific.

This PR adds a condition before applying the filter.

## Test instructions

Checkout and `make client`. The block selector should show normally for any other block and empty for the Feedback Button block.


CGastrell:
The PR doesn't address the issue

PR #150, opened 2021-06-01
Revert poll fixes investigation

CGastrell:
This PR removes some commits as part of a research. #146 is actually a revert of #142, so the only real revert is #135

Commits removed:

  * 6d2225f Fix poll styling issues (#135)
  * cc45e74 Add filter to control transform behavior (#142)
  * 00cc774 Remove the filter as it either turns problematic or doesn't solve the issue at all (#146)


CGastrell:
This has been deployed for a while, no changes were noted regarding our Poll block stats. Closing.

PR #154, opened 2021-07-12
Add a widget-editor compatible interface for the feedback block

ice9js:
This patch will make the feedback block render inline whenever used inside the widget editor.  
Unfortunately we can't have it rendering 'anywhere' as then it's impossible for anyone to tell what area it's actually in. This way it'd also theoretically be possible to have more than one in different areas. I did add a banner on top of the block explaining how it works as well as a subtle border to better imitate where it will be placed on the page.

Screen Shot 2021-07-12 at 11 52 39 PM

As I was fiddling with the styling for the buttons a lot to get this done, I also fixed two other issues involving 'overkill borders' on vertically centered buttons and appending spaces to the button text in Firefox.

Solves c/8P2J5FRq, c/IOYbOdca and c/P6GMxhA7.

# Testing

You'll need to setup a test site with the widget editor enabled in order to test it.

- Make sure the block displays correctly inside the widget editor and its functionality is preserved.
- Make sure the tweaks haven't affected the regular _post_ version of the block.

### DIFF
--- a/client/blocks/feedback/attributes.js
+++ b/client/blocks/feedback/attributes.js
@@ -108,4 +108,8 @@ export default {
 		type: 'boolean',
 		default: false,
 	},
+	isWidget: {
+		type: 'boolean',
+		default: false,
+	},
 };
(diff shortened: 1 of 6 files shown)


CGastrell (approved):
Looking good so far

CGastrell (approved):
Works!

PR #155, opened 2021-07-14
Default to site URL on source_link prop when no permalink is returned

CGastrell:
This PR addresses an issue when Feedback Button block is used on FSE: the permalink is empty.

Depends on #154 and D64087-code

## Test instructions
Checkout and `make client`. Use a FSE enabled site and create a Feedback Button as a widget. Visit your dashboard and check that the created block shows the URL where it was created.


ice9js (approved):
Works for me for NPS and Feedback... and not quite testable for polls like you mentioned.
:shipit:

PR #156, opened 2021-07-16
Update dockerfile

CGastrell:
This PR updates the Dockerfile setup as xenial is not able to pull all the needed packages anymore.

## Test instructions
Checkout and make sure you clean up the old build/images:
```bash
make docker_down
```
List your docker instances to make sure it's not running.

```
docker ps -a
```

If any of the involved instances is running, stop it with `docker stop [container ID]`, then remove them with `docker rm [container ID]`

Remove images:
```
docker image prune
```

The mysql image can be left alone, the rest should be removed via `docker image rm [image ID]`. Some `rm` commands might fail due to dependant/child images. Try different removal order to address this.

Finally, create the docker again:
```
make docker_build
```
and run it with `make docker_up`

You local instance should be running and accessible at http://localhost:8080

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,9 +1,10 @@
-FROM ubuntu:xenial
+FROM ubuntu:groovy
 
 VOLUME ["/var/www/html"]
 
 ENV LANG en_US.UTF-8
 ENV LC_ALL en_US.UTF-8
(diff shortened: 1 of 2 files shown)


digitalwaveride:
Unfortunately ran into the php issue again, during 
`make docker_build`

image

digitalwaveride (approved):

PR #157, opened 2021-07-16
Fix feedback block editor overlay on the full site editor

ice9js:
This patch fixes the issue with the feedback block editor overlay being offset incorrectly inside the full site editor by checking if the block is rendered inside an iframe.

Before:

Screen Shot 2021-07-17 at 12 41 40 AM

After:

Screen Shot 2021-07-17 at 12 42 03 AM

# Testing

Make sure you're testing on WP `trunk` and select a theme which has FSE support.
Verify that when adding the feedback block in side of the full site editor, the overlay is rendered correctly when the block is selected.
Similarly, verify that the overlay still displays correctly on the regular poll editor which doesn't have the iframe.

### DIFF
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -202,6 +202,20 @@ const EditFeedbackBlock = ( props ) => {
 	}, [ attributes.header, popover.current, isSelected ] );
 
 	useLayoutEffect( () => {
+		if (
+			triggerButton.current &&
+			triggerButton.current.ownerDocument !== document
+		) {
+			setOverlayPosition( {
+				bottom: 0,
+				left: 0,
+				right: 0,
+				top: 0,
+			} );
+
+			return;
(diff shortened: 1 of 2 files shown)


CGastrell:
I'm having this issue where I inserted the feedback button block on the body of the page, yet the footer takes precedence on the z-index. So, when hovering, the footer will display this outline and if I click, of course, the focus goes into the footer, not the block.

image

CGastrell (changes requested):
Check on the above comment to see some weird behavior (actually not that weird)

ice9js:
I bumped up the z-index value - since we're positioning it manually anyway, the old comment doesn't apply anymore.

digitalwaveride (approved):
Worked for me as advertised!
image

🚀

CGastrell (approved):
Working fine now :shipit:

PR #163, opened 2021-08-06
Test: make button style default

CGastrell:
This PR explores the idea of making Buttons style the default style.

## Challenges

Both styles have to be [dynamically added and removed](https://github.com/Automattic/crowdsignal-forms/commit/a6b9eecbe6886ebdbf065f44779ab4f934e05e4d) upon availability (Buttons style should only be available when NOT in multiple choice option)

Otherwise, when switching to Multiple Choice mode a "double default style" can be seen, not sure why.

This other challenge I couldn't figure out without making a huge change: for the style to be properly applied by default the CSS class detection has to be inverted and applied by default. This causes ALL existing polls that have been published with the default style (Radio buttons) to now show as Buttons.

I'm quite sure there must be a simpler workaround for the latter, this is just a quick exploration PR.

## Test instructions
Checkout and `make client`. Adding a poll on a post should show the Buttons style selected and applied by default.
Check polls that are created without selecting any style, those should render with Buttons style. This might raise problems.

### DIFF
--- a/client/blocks/poll/index.js
+++ b/client/blocks/poll/index.js
@@ -59,8 +59,8 @@ export default {
 	},
 	styles: [
 		{
-			name: 'default',
-			label: __( 'Default', 'crowdsignal-forms' ),
+			name: 'buttons',
+			label: __( 'Buttons', 'crowdsignal-forms' ),
 			isDefault: true,
 		},
 	],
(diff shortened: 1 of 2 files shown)


digitalwaveride (approved):
That was fast Christian! 🚀 Thank you!
Worked for me! Buttons showed by default!

image

One question though regarding the sidebar setting for Styling:
There is still the "default" setting, the "bullet point list" setting. Confusing for users?
image

CGastrell:
This does not work as it changes all already created polls. Closed in favor of #189

PR #168, opened 2021-08-31
Fix new interface package compatibility

CGastrell:
This PR adds a simple check to see if `isFeatureActive` is available and prioritize its use over `getPreference( 'features' )`.

Our use of `getPreference( 'features' )` aimed to keep a watch on `fullscreenMode` but since https://github.com/WordPress/gutenberg/pull/34154 this seems not available anymore.

## Test instructions
Checkout and `make client`. Open the developer console and insert a Feedback Button on a post. There should be no errors regarding the block insertion.


digitalwaveride (approved):
Error is gone 👍 
image

PR #169, opened 2021-08-31
Fix video size on setup page

CGastrell:
This PR makes video iframe dimension-less and adds CSS rules for it to fill `100%` its container.

## Test instructions
Checkout and visithttp://localhost:8000/wp-admin/admin.php?page=crowdsignal-forms-setup

The video should now be big to fill the wrapped content


digitalwaveride (approved):
🚀

PR #170, opened 2021-08-31
Fix toolbar remaining behind the feedback overlay

CGastrell:
This PR changes the interface skeleton popover z-index: force 101 to not be occluded by our overlay

## Test instructions
Checkout and `make client`. Insert a Feedback Button on a post.

**The first time the toolbar is rendered, it will remain behind the overlay. So far, this is unavoidable due to the mechanics behind the toolbar rendering process.**

De-select the block and select it again: See that the Toolbar is now functional and not behind the overlay


digitalwaveride (approved):
Works as advertised! Let's see if we can find a fix for the last glitch when Kuba is back! 

🙏 Thanks Christian!

PR #172, opened 2021-09-13
Fix toolbar/feedback overlay interactions

ice9js:
This patch reverts Automattic/crowdsignal-forms#170 in favor of a more concise fix for the block toolbar being overlapped by the feedback button's overlay.

The root issue was the z-index property on the block wrapper, which I believe dates back to the days when the overlay would span the entire window and be overlapped by the UI.  
As we later implemented proper logic for sizing the overlay so it only ever covers the content part of the editor, I don't think this value need to be so high. Setting it to `1` makes our block appear in front of other _normal_ blocks while not covering up the toolbar.

While at it, I added an additional fix for the overlay not being resized correctly whenever the sidebar is toggled.

# Testing

- Add a feedback block to your page.
- Trigger the feedback block, the toolbar should still appear in front of the overlay and the overlay should cover the content.
- Add some more regular blocks to the page and test if the overlay covers them when the feedback button is active.
- Try toggling the sidebar and validate that the feedback button and overlay are resized/repositioned correctly.

### DIFF
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -46,8 +46,8 @@ const EditFeedbackBlock = ( props ) => {
 
 	const {
 		attributes,
-		activeSidebar,
 		isFullscreen,
+		isSidebarActive,
 		fallbackStyles,
 		isSelected,
 		setAttributes,
@@ -68,6 +68,7 @@ const EditFeedbackBlock = ( props ) => {
 	} = attributes;
 
 	const [ margin, setMargin ] = useState( {} );
+	const [ sidebarChanged, setSidebarChanged ] = useState( false );
 
 	const widgetEditor = useMemo( isWidgetEditor, [] );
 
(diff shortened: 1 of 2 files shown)


digitalwaveride:
Worked for me at first as advertised!
Tested the block in relation with basic other blocks...
image

... as well as testing it with the right settings sidebar...
image

Things worked as expected, only when testing the left block library sidebar I got these two little hiccups:
 
1)
image

This happens for me on Firefox, when:
- Feedback block is added
- Left block library sidebar is expanded
- Click on Feedback block

2)
image
image

When the button block is placed on the left side, expanding the left block library sidebar is overlapping/hiding the feedback button.

ice9js:
I pushed some more changes to fix the issue with the inserter and noticed the same thing was happening with the list view as well - they should all work alright now.

CGastrell (approved):
I knew there was some better way to do this :D 

LGTM!

PR #173, opened 2021-09-15
Fix Feedback Button block on Block Widget editor.

yansern:
## Description

Fixes the `isWidgetEditor` checking.

## Test Instructions

1. Go to Customize > Widgets.
2. Add "Feedback Button" block.
3. It should not show the error 'This block has encountered an error and cannot be previewed.' and cannot be previewed'.

Fixes https://github.com/Automattic/wp-calypso/issues/55808

### DIFF
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -214,6 +214,10 @@ const EditFeedbackBlock = ( props ) => {
 	}, [ attributes.header, popover.current, isSelected ] );
 
 	useLayoutEffect( () => {
+		if ( isWidgetEditor ) {
+			return;
+		}
+
 		if (
 			triggerButton.current &&
 			triggerButton.current.ownerDocument !== document
--- a/client/blocks/feedback/edit.scss
+++ b/client/blocks/feedback/edit.scss
@@ -51,7 +51,8 @@
 	}
 }
 
-.editor-styles-wrapper {
+.editor-styles-wrapper,
+.wp-customizer {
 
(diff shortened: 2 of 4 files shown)


ice9js (changes requested):
The updated condition works, thank you!  
There's a few more issues we need to address, also `npm run lint:js` is failing.

The block still comes up with _this block cannot be rendered_ message. To fix it, the following condition needs to be added [here](https://github.com/Automattic/crowdsignal-forms/blob/master/client/blocks/feedback/edit.js#L217):

```
useLayoutEffect( () => {
	if ( isWidgetEditor ) {
		return;
	}
	...
```

I also noticed another issue, perhaps more significant... The block is still rendered with incorrect positioning and missing styles. It seems that the customizer is missing `.editor-styles-wrapper` which I believe is what most editor style overrides are supposed to be based on, not just in our block. Hence this is maybe something that needs to be addressed in core?

Otherwise we'll need to find an alternative selector for [these styles](https://github.com/Automattic/crowdsignal-forms/blob/master/client/blocks/feedback/edit.scss#L54).

Screen Shot 2021-09-15 at 4 16 07 PM

On a side note, since the customizer sidebar _is immune_ to any theme styles hence the block has a different appearance in the sidebar than on the preview but maybe that's not an issue because well... there's the preview?

yansern:
Thanks @ice9js. I'm back and I'm looking at this now.

yansern:
Alright, CSS fixes are also in now.

image

ice9js (approved):
Nice! Works for me, thanks again for picking this one up ;)

PR #178, opened 2021-09-27
Move menus into Settings

donnchawp:
This is a dupe of #174 which was reverted.

The "Getting Started with Crowdsignal" menu is a temporary menu until we move it into the main settings page.

PR #179, opened 2021-09-27
WIP: Move the "Getting Started" process into the Settings page.

donnchawp:
Not ready yet - needs changes to text and review/update of process.


donnchawp:
This patch also includes the code from #178, moving menus around, so that PR can probably be ignored.

digitalwaveride (approved):
Navigation is where it should be 👍 !
The disconnect/connect process works! 👍 

In the next step content and styling need just some more finetuning.

PR #181, opened 2021-09-28
Release version 1.5.9 bump

CGastrell:
This is a release PR for version 1.5.9 after #180 bugfix

## Test instructions

* Bugfix: Set the priority of withFixedPosition hook on BlockListBlock to 1 (#180)

This is a AT related bug where some other plugin would take over our block's inline styles.


ice9js (approved):
LGTM!

PR #186, opened 2021-10-27
Fix warnings and skips

CGastrell:
This PR changes some data validation with more reliable checks:

- use `empty` instead of trying to access array index directly
- skip block processing if there's no poll ID present (poll ID should be acquired while editing and available at the time of saving)

## Test instructions
These are mere code improvements, this PR should not pose significant changes unless something is broken on the frontend


ice9js (approved):
>These are mere code improvements, this PR should not pose significant changes unless something is broken on the frontend

Famous Last Words.
Looks good to me 😉

PR #187, opened 2021-10-29
Attempt to track failed connections

CGastrell:
This PR adds a call to our track events to try and solve the disconnection issue happening on isolated cases.

When the connection fails and the notice is shown, try and enqueue a call to the event system.

## Test instructions

Checkout and run `make client`. Force a disconnection by un-setting your Crowdsignal API KEY from the settings.
Create a post and insert a Poll block.
- on self-hosted sites there should be no errors, but no tracking either
- on a sandboxed site you'll need to force the disconnection by deleting the API KEY manually, then try and insert a Poll block on a post

After a couple of minutes you should be able to see the logged event: `crowdsignal_connection_failed`

### DIFF
--- a/client/components/connect-to-crowdsignal/index.js
+++ b/client/components/connect-to-crowdsignal/index.js
@@ -3,11 +3,14 @@
  */
 import { Button } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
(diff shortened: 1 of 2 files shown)


ice9js (approved):
🤞

PR #189, opened 2021-11-05
Default polls to "Button" style

CGastrell:
This PR registers a variation and makes it default with the button style class.

Thanks @ice9js for the tip on this one

## Test instructions
Checkout and run `make client`. Create a post and insert a Poll block. The Poll should be on its "button style" by default.
Check older Poll blocks either on the same post or other posts, those should remain as they were.

### DIFF
--- a/client/blocks/poll/index.js
+++ b/client/blocks/poll/index.js
@@ -63,5 +63,18 @@ export default {
 			label: __( 'Default', 'crowdsignal-forms' ),
 			isDefault: true,
 		},
+		{
+			name: 'buttons',
+			label: __( 'Buttons', 'crowdsignal-forms' ),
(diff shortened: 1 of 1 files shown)


digitalwaveride (approved):
Made so my day! Thank you guys!!! 🙏 

Worked for me! Also old polls kept their old design.

One little thing... not sure if this is doable:
The "List style" is still called "default" in the sidebar even though it is not default anymore.
image

Any chance we could rename this to "List" and "Buttons" ... the default name was always strange... thanks!

CGastrell:
> Any chance we could rename this to "List" and "Buttons" ... the default name was always strange... thanks!

I changed the label to "List"

PR #191, opened 2021-11-17
Rework code

CGastrell:
This PR fixes some small issues with the code and variable naming

* 996cbd1 get_user_code never returns strict false, mostly empty strings
* 19d865c auto-draft is the post status upon creation. Skip any synchronization on this status, it's triggering a useless request on EVERY POST CREATION
* fb8a60d Var rename: api_authenticator is the abstraction for auth_provider, less confusing this way
* 0085fa2 A deprecation notice is taking place on filter usage `block_categories` in favor of `block_categories_all`


jcheringer (approved):
Thanks @CGastrell. LGTM!

PR #193, opened 2021-11-18
Broken connection handling

CGastrell:
This PR is here up for discussion.

When the connection fails to get a valid user/userCode, it will trigger a `crowdsignal_forms_failed_connection` action.

This will allow to take action whenever we need to. Currently no action is taken on self-hosted sites.

Ref: D70389-code


donnchawp (approved):
Looks good!

PR #195, opened 2021-11-23
Update and delete the dashboard login details too

donnchawp:
If this plugin is installed first then make sure the older plugin can
immediately login with the same login credentials.


CGastrell (approved):
Looks fine.

PR #196, opened 2021-11-23
Let the dashboard plugin take care of menus.

donnchawp:
If the other Crowdsignal plugin is installed then it will take care of
connecting/disconnecting from Crowdsignal.
This patch also removes the Getting Started page so it should be merged
*after* https://github.com/Automattic/crowdsignal-forms/pull/179 is
merged.

Test
1. Install the Crowdsignal Plugin from https://github.com/Automattic/crowdsignal-plugin
2. Install this plugin with the add/menu_refresh branch.
3. Enable the Crowdsignal plugin. You should see Settings->Crowdsignal at options-general.php?page=crowdsignal-settings
4. Enable the CS Forms plugin. It shouldn't create a new Crowdsignal menu.


digitalwaveride (approved):
🚀 as advertised!

PR #197, opened 2021-11-23
Some fixes along the way

CGastrell:
This PR is part of the issue tracking regarding our connection problems.

* Rollback usage of `block_categories_all` as it's bringing problems with our builds. Deprecation warning will have to wait
* Use `core` store to get the user ID
* Pass error message as argument when invoking failed connection action
* Abort process if user ID is `0` and, most importantly, DO NOT SET THE TRANSIENT if that is the case

## Test instructions
This all is mostly backend fixes, not much to test but a usage of all the blocks and an eye on error logs / console would help.


donnchawp (approved):
Looks good!

PR #200, opened 2021-11-25
Fix transient and empty user

CGastrell:
This PR changes where we cache for requests and removes the restriction for non logged in users.

* Allow user ID `0`/`null` to perform the request, credentials should be stored on the blog options
* Remove transient on get_account_info, we can't validate a unique-ish key there
* Place transient on auth headers filter, salt it with API KEY
* Fix an issue with a filter handler (dupe method from setup to settings class)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.cache
 .idea
 *.code-workspace
 vendor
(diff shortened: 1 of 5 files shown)


donnchawp (approved):
Looks good!

PR #208, opened 2022-01-18
Fix CSS so it doesn't interfere with Jetpack

donnchawp:
The CSS in the new menus interferes with Jetpack so we need to change our CSS.

### DIFF
--- a/includes/admin/admin-styles.css
+++ b/includes/admin/admin-styles.css
@@ -15,16 +15,71 @@
 	font-weight: 700;
 }
 
-#jp-plugin-container .wrap {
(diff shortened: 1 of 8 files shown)


CGastrell:
Looking good, I'm gonna build from this branch and try it out on WP

digitalwaveride (approved):
With Jetpack installed, didn't see any issues here:

image

🙏 Thanks Donncha!!!

CGastrell (approved):
Didn't see any issues after testing on WP (as expected since we never reach the setup pages there)

PR #212, opened 2022-02-14
Remove templateLock

CGastrell:
This PR removes the `templateLock` restriction on Vote block innerBlocks. The directive not only prevents insertions, but also removals. This goes against what we state on the docs about "removing the thumbs down vote is easy".

If we're not happy with this change, we can always put up a new "Thumbs Up" block, somewhat similar to the Applause block.

## Test instructions
Checkout and run `make client`. Edit a post on your docker and insert a Vote block. You should be able to remove the votes individually.


digitalwaveride (approved):
Works as advertised!

Also not seeing there the "border glitch" as discussed on Slack.
Which is good! 😁 But we should double check after release! 👍

PR #222, opened 2022-05-13
add/crowdsignal-embed block

TheSteveK:
This is a new block based off the core embed block specially designed to embed Crowdsignal surveys.  Links and buttons have been tailored to Crowdsignal surveys.  The block will also only accept known Crowdsignal domains to avoid confusion.

The core functionality of the embed block largely remains the same and the final result is output to the format from the embed block so the native embed process takes over the duties of displaying the embed in the published page.

There is a limitation in the core embed block where the preview only renders to a certain height in the editor, that also exists in this block.

Testing instructions:
Checkout branch and build the plugin.

Verify "Survey" item exists in the block library and shows up when you search for a block via typing / or hitting the inline + button

Try embedding a non-crowdsignal URL like YouTube, verify the error
Try putting in random garbage text, verify the error
Try a valid Survey URL from the survey's sharing settings, confirm the embed preview loads
Preview the page, verify the embed shows up.


digitalwaveride:
Nice Steve!!! So great to see this block! 🥳 

Embedding functionality worked for me! 👍 


### Some design finetuning:

This is what I currently see:

image

This is the [Figma design](https://www.figma.com/file/Y179jGRPQMN4UjydcQ7784/Survey-Embed-Block?node-id=60%3A197):

image


- [ ] The block icon is mixed up, that one is the quiz icon

- [ ] The block description is still missing

image

- [ ] The text above the input field: "Paste a link to the survey you want to display on your site." is missing.
- [ ] The link text below the input field should be more "proactive -> create!": Create a new survey
- [ ] The spacings need to be finetuned
image

Please feel free to copy paste stuff from the basic Crowdsignal embed block:
image


- [ ] The Crowdsignal logo can be a bit smaller -> 40px


---

And one additional thing... let's discuss if it is a bigger issue to implement:


### The settings sidebar sections are still missing:

The reason why we need these links is that users might forget where they can find the results etc... so even if these links just point to the dashboard, that should be helpful.


image


We are close Steve 🚀 ! Thank you!

digitalwaveride (changes requested):
see comment above

TheSteveK:
@digitalwaveride  I think I got everything addressed in the latest commit.  A couple of notes.  The sidebar description and the description that shows in the block picker are the same so I went with the text you provided.  

Also, we already have a component we use in our other blocks that shows that upgrade message.  I used that so the formatting is a bit different than what your figma shows.  Also note, to see it, you'll need to change yourself to a free account.

digitalwaveride:
@TheSteveK hmn... can't see the changes... but also can't see commits in the activity feed here... 🤔

client/blocks/cs-embed/embed-preview.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * WordPress dependencies
+ */
+import { SandBox } from '@wordpress/components';
+import { useBlockProps } from '@wordpress/block-editor';
+
+export default function EmbedPreview( { html } ) {
+	// const html = '<p> this is placeholder html </p>';
```
jcheringer, line 8: Could we remove this comment? 🙂

client/blocks/cs-embed/sidebar.js
```diff
@@ -0,0 +1,82 @@
...
+			>
+				<div>
+					Edit your { typeTextPlural } on{ ' ' }
+					<a
+						href={ dashboardLink }
+						target="_blank"
+						rel="external noreferrer noopener"
+					>
+						Crowdsignal.com
+					</a>
```
jcheringer, line 34: When creating a concatenated string, to make it translatable, there is this helper [createInterpolateElement](https://github.com/WordPress/gutenberg/tree/trunk/packages/element#createInterpolateElement) we need to use.
If you need some examples on how to use it, you can find some [here](https://github.com/search?q=org%3AAutomattic+createInterpolateElement&type=code).

Another thing I didn't understand well is why we need the `typeTextPlural` as an attribute/variable. 
Couldn't we just inline the string?

client/blocks/cs-embed/sidebar.js
```diff
@@ -0,0 +1,82 @@
...
+			>
+				<div>
+					{ __( 'Manage results on', 'crowdsignal-forms' ) }{ ' ' }
+					<a
+						href={ dashboardLink }
+						target="_blank"
+						rel="external noreferrer noopener"
+					>
+						Crowdsignal.com
+					</a>
```
jcheringer, line 57: The same situation to use the `createInterpolateElement` helper.

client/blocks/cs-embed/toolbar.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { ToolbarButton, ToolbarGroup } from '@wordpress/components';
+import { BlockControls } from '@wordpress/block-editor';
+// import { edit } from '@wordpress/icons';
```
jcheringer, line 7: Could we remove the comment?

client/components/icon/quiz.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+function Survey() {
```
jcheringer, line 3: What about changing the function name to comply with the filename?

jcheringer (changes requested):
Thanks for taking care of this Steve.
I added a few comments that could help us keep the code simpler and/or consistent with what we have on our code base.
Please let me know if you have any questions.

TheSteveK:
I've commited changes based on your request @jcheringer  give it another spin :)

jcheringer (approved):
LGTM!

PR #223, opened 2022-05-31
Fix visibility for top bar tabs when Hello Dolly is active

wojtekn:
The styling for Hello Dolly plugin content introduced in the Crowdsignal Forms plugin breaks the top menu navigation in WP when both Crowdsignal forms and Hello Dolly plugins are installed - it hides tabs that are part of the top bar:

Screen Shot 2022-05-31 at 15 40 05

This PR fixes https://github.com/Automattic/wp-calypso/issues/60361

## Test instructions

1. Log in to the test site's WP Admin 
2. Ensure that the Hello Dolly plugin is installed and active
3. Ensure that the Crowdsignal Forms plugin is installed and active
4. Navigate to "Plugins" page

The expected behavior is to have top bar tabs visible like shown on the screenshot:

Screen Shot 2022-05-31 at 15 39 53

### DIFF
--- a/includes/admin/admin-styles.css
+++ b/includes/admin/admin-styles.css
@@ -82,21 +82,6 @@
   color: #888;
 }
 
-.wp-admin #dolly {
-	float: none;
-	position: relative;
-	right: 0;
-	left: 0;
-	top: 0;
-	padding: .625rem;
-	text-align: right;
-	background: #fff;
-	font-size: .75rem;
-	font-style: italic;
-	color: #87a6bc;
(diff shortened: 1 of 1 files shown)


vykes-mac:
Tested this and it work as expected, I realise though that's issue seem to be caused by the `position: relative` in the css class. I wonder if it would be better to maintain the original styling by just removing the `position` instead of the entire class?

vykes-mac:
Ah I see 👍🏾

vykes-mac (approved):

PR #227, opened 2022-06-29
Build Tools: Add `allow-plugins` to `composer.json` for v2-compat

danielbachhuber:
See https://getcomposer.org/doc/06-config.md#allow-plugins

The change is for `composer` v2 compatibility. `composer v1` won't be impacted by this change.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,5 +10,11 @@
 		"wp-coding-standards/wpcs": "2.2.1",
 		"wp-cli/wp-cli-bundle": "2.4.0",
 		"psy/psysh": "^0.10.7"
+	},
+	"config": {
+		"allow-plugins": {
+			"cweagans/composer-patches": true,
+			"dealerdirect/phpcodesniffer-composer-installer": true
+		}
 	}
 }


danielbachhuber:
@CGastrell FYI - this is still valid.

CGastrell:
> @CGastrell FYI - this is still valid.

Fantastic, it's just that I don't know what this does or what it improves, can you clarify on it?

CGastrell:
I see. Seems fine, where did you get those "allowed plugins" from? Did you run into trouble getting our docker instance, while installing?

danielbachhuber:
> Seems fine, where did you get those "allowed plugins" from?

The `allowed-plugins` are either direct or child dependencies:

https://github.com/Automattic/crowdsignal-forms/blob/f2989e34c0618f829d7f8d26f9383d61bb082605/composer.json#L9

https://github.com/Automattic/crowdsignal-forms/blob/f2989e34c0618f829d7f8d26f9383d61bb082605/composer.lock#L6086

> Did you run into trouble getting our docker instance, while installing?

Not sure how this is related to docker?

CGastrell (approved):
Thanks a lot @danielbachhuber

danielbachhuber:
@CGastrell You're welcome!

PR #230, opened 2022-07-05
add/quiz variation to CS Embed block

TheSteveK:
As promised, this PR adds a variation to our Embed block to add a quiz block via CS Forms using the block variation method. 
Screen Shot 2022-07-05 at 9 41 23 AM
Screen Shot 2022-07-05 at 9 41 35 AM
Screen Shot 2022-07-05 at 9 41 55 AM

Testing:

Apply this branch and check that Quiz appears in the block picker/block search.  Check the text to ensure it says Quiz where it should.

Embed a quiz and verify it renders like Surveys do
Ref https://href.li/?351-gh-Automattic/crowdsignal


client/blocks/cs-embed/edit.js
```diff
@@ -105,7 +105,7 @@ const EmbedForm = ( { attributes, setAttributes } ) => {
 			) : (
 				<Placeholder
 					icon={ CSLogo }
-					label={ __( 'Survey Embed', 'crowdsignal-forms' ) }
+					label={ placeholderTitle }
```
ice9js, line 108: This isn't necessarily wrong but I do wonder if it's really the best way to keep these things inside block attributes 🤔.  
I think a potentially better way would be to create a separate component for the placeholder and define a map of different props to use based on what the current variation is.

Really, it kind of works the same way except it's a separate object and doesn't pollute `attributes` with _random data_ that's really not used for anything ever again once the block is created.
I'm not sure it's worth changing this now but do keep it in mind in the future :)

ice9js (approved):
Looks good to me!

PR #233, opened 2022-07-08
fix/add missing ref links to quiz variation

TheSteveK:
I forgot the referral links for the quiz variation.  This adds them in.

Testing.  Load it up, add a quiz block, check that the links in the block and sidebar have ?ref-quizembedblock at the end of them.


TheSteveK:
No idea, but that's what Thomas asked for :)

PR #236, opened 2022-07-19
Update development environment

ice9js:
This patch makes a few tweaks to our development environment to make it more friendly to use:

- Bumped up `PHPUnit` to `^9` which adds compatibility for PHP 8 and up. (v7 was restricted to PHP 7.x)
- Updated `make install-node` to use `npm ci` instead of `npm install` forcing it to actually respect `package-lock.json`.

# Testing

Run `make all` and make sure everything installs and builds correctly. Verify on a local test site.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ install: install-node install-php
 
 # Install Node dependencies
 install-node:
-	npm install
+	npm ci
 
 # Install PHP dependencies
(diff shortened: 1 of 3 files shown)


CGastrell (approved):
LGTM! I also added the specific phpunit version to the installation script (Dockerfile) on #237

PR #239, opened 2022-07-19
Remove setup page

CGastrell:
Seems like this is a remnant of previous version of "settings" page

Need to verify, but seems it's not used anymore

NOT READY FOR MERGE


CGastrell:
Outdated and doesn't make sense anymore

PR #252, opened 2023-03-24
Fix rendering on question and note

CGastrell:
This PR removes some old usage of RawHTML

TBD


xknown (approved):
LGTM.

I think we probably need to remove the use of RawHTML in some other places https://github.com/Automattic/crowdsignal-forms/search?q=RawHTML as it might result in similar XSS issues.

PR #255, opened 2023-04-06
Remove RawHTML use from the blocks

ice9js:
Followup on #252.

This patch removes remaining instances of `<RawHTML>` from the plugin blocks and replaces it with `decodeEntities()`.

#Testing

This patch affects the feedback and NPS blocks. They should still work as expected and you shouldn't be able to embed XSS payloads inside either of their fields.


CGastrell:
Beware of the header edit with decodeEntities, I think we allow line breaks there

CGastrell:
> Beware of the header edit with decodeEntities, I think we allow line breaks there

@ice9js 
image

CGastrell (changes requested):
A minor hiccup with the variable and a couple other places where we could use the same fix

PR #258, opened 2023-04-13
fix block spacing for better theme compatibility

CGastrell:
This PR changes some margin/paddings and display styles so buttons and blocks are more theme change tolerant.

Tested on themes 2020, 2021, 2022, 2023, Astra and Seedlet

## Test instructions

Publish a post with one of each Poll (default/button and MC variations), Applause, Vote, Feedback and NPS bocks.

Compare editor and published UI/layout to see they match as much as possible. An issue with border-radius remain as old themes used to provide those in a different manner.

### DIFF
--- a/client/blocks/poll/edit.scss
+++ b/client/blocks/poll/edit.scss
@@ -33,6 +33,7 @@
 
 		&.wp-block-button__link {
 			cursor: text;
+			display: block;
 
 			&:hover {
 				opacity: inherit;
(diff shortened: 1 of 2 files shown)


digitalwaveride (approved):
Thanks so much Christian for looking into this! 🙏 🙏 🙏 

I tested on themes 2020, 2021, 2022, 2023 and Seedlet.

Button styling on all these themes was improved, and as expected!

image

As discussed on Slack, I had on my local sandbox a little issue with the styling of the "notes section" regarding font-size, on themes 2022 and 2021, but that seemed only on my local machine.

PR #259, opened 2023-04-13
Fix line jumps on polls

CGastrell:
This PR applies the split/join strategy to fix line jump rendering on poll question and note after our RawHTML deprecation.

## Test instructions

Publish a poll. Add line jumps on question and note. Published post should reflect the line jumps instead of rendering `<br>`


ice9js (approved):
👍

PR #262, opened 2023-04-26
Update dependencies for node 18.13.0

CGastrell:
Janitorial: this PR updates packages and node version to 18.13.0

Includes an update to the GH action to generate builds.


Makefile
```diff
@@ -3,26 +3,22 @@
...
-install: install-node install-php
+install: install-pnpm install-node install-php
 
 # Install Node dependencies
 install-node:
-	npm install
+	pnpm install
+
+install-pnpm:
+	npm install -g pnpm@7.27.1
```
ice9js, line 13: This is the only thing I'd say is a strong blocker. I'd much rather have the extra line of instructions in our readme or something than having `make` implicitly change my global environment without any sort of warning!

ice9js (changes requested):
All good, minus the `install-pnpm` step on the `Makefile`. I'd say let's just drop it to be honest.

Makefile
CGastrell: Mmm... fair enough

CGastrell: It's on the development-environment.md, I'll remove it from the Makefile though

ice9js (approved):
Thanks Christian!

PR #263, opened 2023-04-26
Optimize frontend components

CGastrell:
This PR attempts to remove, as far as possible, dependencies on @wordpress/components on the frontend components to lower the size of the builds and improve render performance.


client/blocks/feedback/edit.js
```diff
@@ -470,7 +504,24 @@ const EditFeedbackBlock = ( props ) => {
...
+													target="_blank"
+													rel="noopener noreferrer"
+													className="crowdsignal-forms__branding-promote"
+												>
+													{ __(
+														'Hide',
+														'crowdsignal-forms'
+													) }
+												</a>
+											</Tooltip>
```
sgomes, line 523: You may have considered this already, but one possibility would be to extract this to a separate module, to be used for both `feedback/edit.js` and `nps/edit.js`. That would still keep it out of the frontend, and should be handled just fine by the build process.

client/components/nps/index.js
```diff
@@ -55,15 +54,20 @@ const Nps = ( {
...
+						.split( '<br>' )
+						.join( '\n' ) }
 				</h3>
 
 				<button
 					className="crowdsignal-forms-nps__close-button"
 					onClick={ onClose }
 				>
-					<Icon icon="no-alt" />
+					X
```
sgomes, line 70: Suggest using ✖ or ✕ instead, which are a bit more visually distinct from the letter X.

Alternatively, you could embed a tiny SVG here; in general it's a bad practice to do SVG-in-JS, but if it's something tiny that you're using in a single spot, that's fine.

client/components/with-fallback-styles/util.js
```diff
@@ -37,3 +37,81 @@ export const getBorderColor = ( borderNode ) => {
...
+ */
+import fastDeepEqual from 'fast-deep-equal/es6';
+
+/**
+ * WordPress dependencies
+ */
+import { Component } from '@wordpress/element';
+import { createHigherOrderComponent } from '@wordpress/compose';
+
+export const withWordPressFallbackStyles = ( mapNodeToProps ) =>
```
sgomes, line 52: Good call on moving this method here! 👍

Given the way that Gutenberg recommends externs for WP and how big `@wordpress/components` is, it's really not a great idea to have utility methods in there 😕

sgomes (approved):
LGTM! 👍

I left a few comments, and I think this is probably best reviewed by someone that has a deeper understanding of the codebase as well, but the approach and the code look good. 

Thank you again for working on this, @CGastrell! 🙏

client/blocks/feedback/edit.js
CGastrell: Yeah, I think I was mostly dashing and slashing until I got it out, likely suspect to get into a module on its own.

CGastrell: I moved the tooltip to a separate component so to reuse it. Thanks for the suggestion!

PR #265, opened 2023-08-16
Block Sidebar: group settings together

kraftbj:
When using the Poll block, I didn't realize there were more settings regarding the poll itself after the styling panel. Suggesting moving the Answer Settings panel above the styling to sit alongside the other settings panels.

The Feedback and NPS blocks are similar; updated PR to include those.


kraftbj:
Before: the Answer Settings is not visible on a large iMac screen:
2023-08-16 at 9 40 AM

ice9js (approved):
Works as advertised, thanks!

PR #266, opened 2023-08-17
Update @wordpress/scripts and webpack to latest versions

ice9js:
It seems the `@wordpress/scripts` version provides configuration for Webpack 5 while we were still on Webpack 4 making it impossible to build anything.

This patch updates Webpack to version 5 and `@wordpress/scripts` to the latest version.

### Testing

- Verify that all commands from `package.json` still work as expected and bundles are being built correctly.


CGastrell (approved):
LGTM! Tested fine. I did have to use `pnpm` instead of `npm`, but I think we should switch to it anyways. :shipit:

PR #268, opened 2023-11-07
Load blockObserver on DOM ready instead of window load

### LINKED ISSUE #250: Front end poll display on window load
Noticed there's a delay in the polls loading on sites that have longer load times (e.g. sites with ads that will take time to return and delays the full page load time).

Found the reason for this in the source code and here in the repo:
https://github.com/Automattic/crowdsignal-forms/blob/a9bed7b2cb4d1556fecc5890b0d5491502097196/client/lib/mutation-observer/index.js#L12

I did a quick swap for `document.addEventListener ( 'DOMContentLoaded'` in the source code for testing and the polls loaded faster on the FE and didn't notice any issues with loading or interactivity.
If the swap is possible it would greatly help with loading the polls without being dependent of 3rd party elements delaying load time.

I'd do a pull request with the suggestion but I didn't see a contribution guide file?

mae829:
Resolves #250 

On sites with 3rd party elements and dependency scripts (like Ads) the components take a while to load because of the window load event trigger.
This switches the `blockObserver` to load on DOM load/ready for a more optimized delivery and load of the components to the user.

### DIFF
--- a/client/lib/mutation-observer/index.js
+++ b/client/lib/mutation-observer/index.js
@@ -9,7 +9,7 @@ const MutationObserver = ( dataAttributeName, blockBuilder ) => {
 		return blockObserver( dataAttributeName, blockBuilder );
 	}
 
-	window.addEventListener( 'load', () =>
+	document.addEventListener( 'DOMContentLoaded', () =>
 		blockObserver( dataAttributeName, blockBuilder )
 	);
 };
--- a/client/lib/mutation-observer/index.js
+++ b/client/lib/mutation-observer/index.js
@@ -5,7 +5,7 @@ import { render } from 'react-dom';
 import { camelCase, isEmpty, forEach } from 'lodash';
 
 const MutationObserver = ( dataAttributeName, blockBuilder ) => {
-	if ( 'complete' === document.readyState ) {
+	if ( 'complete' === document.readyState || 'interactive' === document.readyState ) {
 		return blockObserver( dataAttributeName, blockBuilder );
 	}
 


client/lib/mutation-observer/index.js
```diff
@@ -9,7 +9,7 @@ const MutationObserver = ( dataAttributeName, blockBuilder ) => {
 		return blockObserver( dataAttributeName, blockBuilder );
 	}
 
-	window.addEventListener( 'load', () =>
+	document.addEventListener( 'DOMContentLoaded', () =>
```
jgcaruso, line 12: Since this is changing to `DOMContentLoaded` we may find ourselves in an inbetween state where `blockObserver` is never called since the condition just above this is only checking for `'complete' === document.readyState`.

Can you modify it to check for both `complete` and `interactive`?

https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState

jgcaruso (changes requested):
Hi there, thanks for contributing this PR! Sorry for long wait time for a review.

I haven't run through a test yet, but I looked at the code and had one suggested change for you to consider. What are your thoughts on it?

client/lib/mutation-observer/index.js
mae829: @jgcaruso code updated to watch either state update

jgcaruso (approved):
Thanks for the contribution!

Code changes look good, and I tested this against a known slow script resource and the poll is rendered right away, before page load fires, and the poll vote works as expected.

Issue #185, opened 2021-10-15
No space between closed poll and thank you messages


danis039:
When a user votes on a poll which has the option "One response per computer" enabled and later when this poll has been closed, the message displayed by default is the following: "This Poll is ClosedThanks For Voting!". As seen in the screenshot below, the messages are shown together with no spacing between them.  

![image](https://user-images.githubusercontent.com/11977521/137531957-0f11dc85-8168-45d4-b0e7-c12e25330b8c.png)

As these messages are not wrapped in `span` elements it's not really possible to style them with CSS.

Any possibility to have a fix for this, please?

Thanks in advance! Loving the plugin :)

bobmatyas:
I don't believe this behavior still exists. It looks like it was addressed on https://github.com/Automattic/crowdsignal-forms/pull/269

Screenshot:

<img width="3674" height="2090" alt="Image" src="https://github.com/user-attachments/assets/66e0f506-d2f6-4bf5-8796-55704327801d" />

PR #269, opened 2023-12-01
Wrap closed banner messaging in span tags

colinswinney:
The Closed Banner messaging on polls does not include line breaks.  This PR would wrap the various messages in `<span>` tags and add additional styles so that each message would be on it's own line. 

See also this issue: https://github.com/Automattic/crowdsignal-forms/issues/185

70e3d8f5ade76eb96a4ec191a7d41b0d

### DIFF
--- a/client/components/poll/closed-banner.js
+++ b/client/components/poll/closed-banner.js
@@ -20,11 +20,16 @@ const ClosedBanner = ( {
 
 	return (
 		<div className={ classes }>
-			{ isPollHidden && __( 'This Poll is Hidden', 'crowdsignal-forms' ) }
+			{ isPollHidden && (
+				<span>{ __( 'This Poll is Hidden', 'crowdsignal-forms' ) }</span>
+			) }
 			{ isPollClosed &&
-				! isPollHidden &&
-				__( 'This Poll is Closed', 'crowdsignal-forms' ) }
-			{ hasVoted && __( 'Thanks For Voting!', 'crowdsignal-forms' ) }
+				!isPollHidden && (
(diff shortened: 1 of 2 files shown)


thedebian (changes requested):
Instead of wrapping messages in span, it'd be better to display a single message, in the following order of priority:
* If it's hidden, then hidden message displayed.
* If not, then if it's closed, closed message displayed
* If not, then if has voted, thanks for voting message displayed
There's no reason to say "thanks for voting" on a closed poll

colinswinney:
That sounds good @thedebian, those changes have been addressed.

client/components/poll/closed-banner.js
```diff
@@ -18,18 +18,18 @@ const ClosedBanner = ( {
 		'crowdsignal-forms-poll__closed-banner'
 	);
 
+	let message = '';
+	if ( isPollHidden ) {
+		message = __( 'This Poll is Hidden', 'crowdsignal-forms' );
+	} else if ( !isPollHidden && isPollClosed ) {
+		message = __( 'This Poll is Closed', 'crowdsignal-forms' );
+	} else if ( !isPollHidden && hasVoted ) {
```
thedebian, line 26:
```suggestion
	} else if ( isPollClosed ) {
		message = __( 'This Poll is Closed', 'crowdsignal-forms' );
	} else if ( hasVoted ) {
```

thedebian: As the first condition is isPollHidden, no need to replicate the condition for the else if

colinswinney: Ahh, good call.  This has been amended.

thedebian (approved):

PR #270, opened 2024-01-23
crowdsignal applause block: Only try to fetch poll data if a pollId exists (php 8.1 warning)

mreishus:
Fixes warnings like:
```
Warning: Undefined array key "pollId" in ...crowdsignal-forms/next/includes/frontend/blocks/class-crowdsignal-forms-applause-block.php on line 76
```
`$attributes['pollId']` could be blank, so only run `get_platform_poll_data()` on it if it exists.

See testing directions on D135692-code. I started there but then realized that file probably belongs to an external repository.


jgcaruso (approved):
Change looks good and tested well on my non-8.1 wpcom sandbox.

mreishus:
Great, thanks!

mreishus:
deployed here: D135692-code

PR #272, opened 2024-03-26
Prevent blocks from being used in the Site Editor

jgcaruso:
There have been a few instances recently where users have been trying to use Poll blocks in FSE and when attempting to vote on the public page, there is a JS error saying an attribute is missing. This plugin may have been created before FSE was available so this scenario was not properly handled.

All of the crowdsignal blocks only support being rendered in a Post context in order for their content to be properly saved since they only get processed on post save hooks.

This PR adds some checks to the block in the editor to try to determine if it is running in an FSE context (no post id) or in a query loop (queryId is available in context). Query loop should not be supported on FSE or in a post because the save hook would always attempt to save against the post that is being saved, not the dynamic post id from the query loop. So the result would be weird.

This PR also adds a check on the rendered page, so if "platform data" (for polls) or surveyId is missing (the data that matches the poll and options to IDs on Crowdsignal) then it won't attempt to render the poll. This is to protect against any existing polls that may have already been created. Hopefully the missing poll on the page would prompt someone to check the editor, see the message, and then not contact support because of a broken poll. This change required a small refactor to cache the data since the "platform data" function is being called twice now, as well as a small bug fix for a weird array with a space in it when data could not be found.

Finally, this PR fixes some issues with the docker container used for local dev / testing.

There is currently an issue with the NPS block prior to this PR where if you open a post for editing that already had an NPS block on it, it fails to render in the editor. I've opened a separate issue to track this https://github.com/Automattic/crowdsignal-forms/issues/273

I also bumped the minimum WordPress version to ensure that we have support for `usesContext` because it seems it was only added at some point in 2020 (WP 5.5?), and current minimum 5.0 version was released in 2018. But having a more modern version will allow us to use more modern features when we are ready to.

**Testing**
1. Apply this PR to get the docker fixes, then run the steps to setup the dev environment. See https://github.com/Automattic/crowdsignal-forms/tree/master/docker#setup
2. Check out `master` again. We want to do these next steps against existing code.
3. On a new post, create a Poll, Vote and Applause block.
   - use the `/poll`, `/vote` and `/applause` blocks, not the embeds
   - also test with the `/nps` and `/feedback` blocks
4. Save the post
6. In FSE, add a Poll, Vote and Applause block anywhere on the homepage not in a query loop
7. In FSE, add a Poll, Vote and Applause block to a query loop's "Post Template"
8. Save the page and view the public pages.
9. The polls should render, but voting should fail with console errors on the homepage
10. Checkout this PR branch again
11. run `make client`
12. refresh the public page, none of the Poll, Vote or Applause blocks should render
13. View the public post page, voting should work as expected
14. View FSE, you should see error messages in replacement of the blocks
15. View the Post editor for your test post. The block should render as expected (with the exception of the nps block as noted in the description above)

### DIFF
--- a/client/components/with-fse-check/index.js
+++ b/client/components/with-fse-check/index.js
@@ -1,0 +1,26 @@
+/**
+ * External dependencies
+ */
+import React from 'react';
+
+import ErrorBanner from 'components/poll/error-banner';
+import { __ } from '@wordpress/i18n';
+
+const withFseCheck = ( Element ) => {
+	return ( props ) => {
+		const { context } = props;
+		const { postId, queryId } = context;
+console.log('fse check')
+		// Prevent block from loading in FSE or a query loop because save handlers don't support those contexts.
+		// - double == instead of triple === used because we need to test for both null and undefined
+		if ( null == postId ) {
+			return <ErrorBanner>{ __( 'Crowdsignal blocks cannot be used outside of a post or page. The Site Editor is not supported.', 'crowdsignal-forms' ) }</ErrorBanner>;
+		} else if ( null != queryId ) {
+			return <ErrorBanner>{ __( 'Crowdsignal blocks are not supported inside a query loop.', 'crowdsignal-forms' ) }</ErrorBanner>;
+		}
+console.log('allowing through')
+		return <Element { ...props } />;
+	};
+};
+
+export default withFseCheck;
--- a/client/components/with-fse-check/index.js
+++ b/client/components/with-fse-check/index.js
@@ -10,7 +10,7 @@ const withFseCheck = ( Element ) => {
 	return ( props ) => {
 		const { context } = props;
 		const { postId, queryId } = context;
-console.log('fse check')
+
 		// Prevent block from loading in FSE or a query loop because save handlers don't support those contexts.
 		// - double == instead of triple === used because we need to test for both null and undefined
 		if ( null == postId ) {
@@ -18,7 +18,7 @@ console.log('fse check')
 		} else if ( null != queryId ) {
 			return <ErrorBanner>{ __( 'Crowdsignal blocks are not supported inside a query loop.', 'crowdsignal-forms' ) }</ErrorBanner>;
 		}
-console.log('allowing through')
+
 		return <Element { ...props } />;
 	};
 };
--- a/README.TXT
+++ b/README.TXT
@@ -1,7 +1,7 @@
 === Crowdsignal Forms ===
 Contributors: automattic
 Tags: polls, forms, surveys, gutenberg, block
-Requires at least: 5.0
+Requires at least: 6.0
 Requires PHP: 5.6.20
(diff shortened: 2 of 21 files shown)


client/components/with-fse-check/index.js
```diff
@@ -0,0 +1,26 @@
...
+import React from 'react';
+
+import ErrorBanner from 'components/poll/error-banner';
+import { __ } from '@wordpress/i18n';
+
+const withFseCheck = ( Element ) => {
+	return ( props ) => {
+		const { context } = props;
+		const { postId, queryId } = context;
+console.log('fse check')
```
gikaragia, line 13:
```suggestion

```

client/components/with-fse-check/index.js
```diff
@@ -0,0 +1,26 @@
...
+		// Prevent block from loading in FSE or a query loop because save handlers don't support those contexts.
+		// - double == instead of triple === used because we need to test for both null and undefined
+		if ( null == postId ) {
+			return <ErrorBanner>{ __( 'Crowdsignal blocks cannot be used outside of a post or page. The Site Editor is not supported.', 'crowdsignal-forms' ) }</ErrorBanner>;
+		} else if ( null != queryId ) {
+			return <ErrorBanner>{ __( 'Crowdsignal blocks are not supported inside a query loop.', 'crowdsignal-forms' ) }</ErrorBanner>;
+		}
+console.log('allowing through')
```
gikaragia, line 21:
```suggestion
```

gikaragia (approved):
Feedback Button crashes on FSE although this doesn't seem related to this PR:
Screenshot 2024-04-05 at 12 44 38

Conceptually, it seems like a weird limitation to not allow users to create polls to templates. As I understand it we do this because we save stuff for polls in post meta (or comment meta). Couldn't we store these in attributes instead? Just asking out of curiosity, I understand that this is too much work to implement at this point.

Looks good for now, just a couple of `console.log`s that should be removed.

client/components/with-fse-check/index.js
jgcaruso: 🤦 thanks for catching that

jgcaruso:
> Conceptually, it seems like a weird limitation to not allow users to create polls to templates.

I agree, having a single poll that exists on every page as part of the template could make sense (someone was trying to do just that in a sidebar which is why I had to fix this).

I think this plugin was created before FSE existed, so it just never considered how to support it. Different hooks would probably need to be created to make that work, and api endpoints changed to support a new set of data for linking the blocks to crowdsignal data. Its probably a big effort as you mentioned.

> Couldn't we store these in attributes instead?

I vaguely remember an issue with storing too much identifying data in the attributes. It may have had something to do with copy/pasting blocks, or just in general keeping things in sync and deciding which end of the integration (WP site or Crowdsignal API) gets to generate the IDs for polls and questions.

Issue #273, opened 2024-03-27
NPS block renders error in editor when re-opening after publish


jgcaruso:
The NPS block is rendering a "This block has encountered an error and cannot be previewed." in the editor when viewed on an existing published post.

The block still works as expected on the published page.

jgcaruso:
Temporarily disabling the Preview button for now in #274 to fix this issue.

PR #274, opened 2024-03-27
disable PostPreviewButton because it is crashing the block on re-renders

jgcaruso:
This PR removes the "Preview" button on the NPS block because its presense is causing the block to blow up in the editor.

Tracking in https://github.com/Automattic/crowdsignal-forms/issues/273

**Testing**
1. First test in `master`
2. Add an NPS block to a post
3. Publish the post
4. refresh the editor
5. The block will show a "This block has encountered an error and cannot be previewed" error
6. Apply this branch
7. run `make client`
8. refresh the editor
9. The block will render

### DIFF
--- a/client/blocks/nps/edit.js
+++ b/client/blocks/nps/edit.js
@@ -152,16 +152,18 @@ const EditNpsBlock = ( props ) => {
 					isDismissible={ false }
 					icon="visibility"
 					componentActions={ [
-						<PostPreviewButton
-							key={ 1 }
-							className={ [
-								'is-secondary',
-								'components-notice__action',
-								'crowdsignal-forms-nps__preview-button',
-								attributes.surveyId ? '' : 'is-disabled',
-							] }
-							textContent={ __( 'Preview', 'crowdsignal-forms' ) }
-						/>,
+						// Temporarily disabled to prevent block render error,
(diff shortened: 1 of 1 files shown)


gikaragia (approved):
Spent some time debugging but couldn't figure out. For some reason can't log anything in the console. Are `console.log`s removed during the build step? Also couldn't figure out if it is possible to enable source mapping to see where the issue is coming from, although I enabled `define( 'SCRIPT_DEBUG', true )`.

Seems like something is messed up in the order `componentActions` are mapped in `EditorNotice` but couldn't figure it out.

jgcaruso:
@gikaragia 

> Also couldn't figure out if it is possible to enable source mapping to see where the issue is coming from, although I enabled `define( 'SCRIPT_DEBUG', true )`.

Yeah I ran into the same issue too. Not sure why the react debug scripts weren't being served with that constant set.

> For some reason can't log anything in the console. Are console.logs removed during the build step?

I'm not sure about this.

> Seems like something is messed up in the order componentActions are mapped in EditorNotice but couldn't figure it out.

Yeah it seemed like something in Gutenberg itself, not necessarily something the block was doing wrong. If we really want to get to the bottom of it, we may need to setup a WordPress/Gutenberg dev environment?

PR #281, opened 2024-04-25
Fix: Made string translatable in html-admin-setup-step-1.php file

DAnn2012:
Fix: Made string translatable in html-admin-setup-step-1.php file

### DIFF
--- a/includes/admin/views/html-admin-setup-step-1.php
+++ b/includes/admin/views/html-admin-setup-step-1.php
@@ -23,7 +23,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 	<div class="crowdsignal-setup__content">
 		<div class="crowdsignal-setup__description">
 				<h1><?php esc_html_e( 'Welcome to Crowdsignal Forms', 'crowdsignal-forms' ); ?></h1>
-				<p><?php echo wp_kses_post( 'To collect and manage responses you need to connect the plugin to <a href="https://crowdsignal.com">Crowdsignal</a>. <br />It will take less than a minute and it’s free.', 'crowdsignal-forms' ); ?></p>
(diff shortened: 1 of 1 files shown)


includes/admin/views/html-admin-setup-step-1.php
```diff
@@ -23,7 +23,7 @@
 	<div class="crowdsignal-setup__content">
 		<div class="crowdsignal-setup__description">
 				<h1><?php esc_html_e( 'Welcome to Crowdsignal Forms', 'crowdsignal-forms' ); ?></h1>
-				<p><?php echo wp_kses_post( 'To collect and manage responses you need to connect the plugin to <a href="https://crowdsignal.com">Crowdsignal</a>. <br />It will take less than a minute and it’s free.', 'crowdsignal-forms' ); ?></p>
+				<p><?php echo wp_kses_post( __( 'To collect and manage responses you need to connect the plugin to <a href="https://crowdsignal.com">Crowdsignal</a>. <br />It will take less than a minute and it’s free.', 'crowdsignal-forms' ) ); ?></p>
```
jgcaruso, line 26: We generally don't like the URLs to be part of the translation, to avoid any potential issues with translators breaking the link.

Would you mind following the pattern used here? https://github.com/Automattic/crowdsignal-forms/blob/e4c2eb37e4898cf68eb0908483f295be647ffafc/includes/admin/views/html-admin-dashboard-teaser.php#L41-L48

jgcaruso (changes requested):
Thanks for the contribution, and sorry for the delay in reviewing it.

Just have a suggestion to make the translation pattern better match how we translate strings with URLs in the rest of the project.

includes/admin/views/html-admin-setup-step-1.php
DAnn2012: Hi @jgcaruso 

I made the changes.

Thanks.

jgcaruso (approved):
Awesome, thanks for the quick changes!

PR #282, opened 2024-06-04
Fix dynamic property

PatelUtkarsh:


jgcaruso (approved):
Thanks for the contribution!

This change gets rid of the dynamic property deprecation warning, as expected.

PR #293, opened 2025-04-25
Add a filter that users can use to prevent redirect on activation.

StevenDufresne:
Not all use cases require that users be redirected to the Crowdsignal settings page. For example, on WordPress.com, we provision a new WordPress installation and install several default plugins. When the user first visits `plugins.php`, they are currently redirected to the Crowdsignal configuration page out of context.

This PR introduces `crowdsignal_forms_allow_activation_redirect` that will allow consumers to set it to `false` programmatically and avoid the redirect when the plugin is activated.

### DIFF
--- a/README.TXT
+++ b/README.TXT
@@ -67,6 +67,9 @@ Compare our [simple and affordable plans](https://crowdsignal.com/pricing/) or t
 
 == Changelog ==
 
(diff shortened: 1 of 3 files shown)


donnchawp (approved):
Thanks, sorry it took so long to get to this!

PR #305, opened 2026-03-26
Refactor withFallbackStyles to remove wrapper div and use iframe-safe APIs

stephdau:
## Summary

WordPress 7.0 moves the block editor into an iframe, which means blocks using
`window.getComputedStyle()` will measure the wrong document. The
`withFallbackStyles` HOC also wraps every consumer in an extra `<div ref>`,
which will conflict with the `useBlockProps()` requirement in Block API v3
(the outermost element must carry block props).

See [Migrating Blocks for iframe Editor Compatibility](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-api-versions/block-migration-for-iframe-editor-compatibility/).

This PR:

- Removes the wrapper `<div ref={this.bindRef}>` from the HOC; passes a
  `fallbackStylesRef` callback prop instead, letting each consumer attach
  the ref to its own root element
- Replaces all `window.getComputedStyle()` calls with
  `node.ownerDocument.defaultView.getComputedStyle()` so style measurement
  works inside an iframe
- Updates all 9 consumers (4 editor blocks + 5 frontend components) to
  attach the new ref

### Files changed (11 files)

**HOC core:**
- `client/components/with-fallback-styles/util.js` — removed wrapper div, iframe-safe getComputedStyle
- `client/components/with-fallback-styles/index.js` — pass through fallbackStylesRef, iframe-safe getComputedStyle

**Editor blocks** (attach `fallbackStylesRef`):
- `client/blocks/poll/edit.js`
- `client/blocks/vote-item/edit.js`
- `client/blocks/nps/edit.js`
- `client/blocks/feedback/edit.js`

**Frontend components** (attach `fallbackStylesRef`):
- `client/components/poll/index.js`
- `client/components/vote/index.js`
- `client/components/applause/index.js`
- `client/components/nps/index.js`
- `client/components/feedback/index.js`

## Testing

1. Build: `nvm use 18 && pnpm build` — no errors
2. Insert poll, NPS, feedback, and vote-item blocks in the editor — verify
   theme color detection still works (fallback styles probe measures correctly)
3. Preview frontend rendering of poll, vote, applause, NPS, and feedback
   blocks — verify theme-matched colors appear
4. No new console errors or warnings related to this change

PR #306, opened 2026-03-26
Migrate all blocks to Block API v3 with useBlockProps

stephdau:
## Summary

Depends on #305.

WordPress 7.0 deprecates Block API versions 1 and 2. All 7 Crowdsignal Forms
blocks default to v1 (no `apiVersion` declared), producing deprecation warnings
and preventing the editor from running blocks inside its iframe.

See [Migrating Blocks for iframe Editor Compatibility](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-api-versions/block-migration-for-iframe-editor-compatibility/).

This PR adds `apiVersion: 3` and `useBlockProps()` to every block:

- **All 7 blocks:** added `apiVersion: 3` to block registration
- **Blocks with `withFallbackStyles`** (poll, vote-item, nps, feedback):
  merged `blockProps.ref` with `fallbackStylesRef` via a callback ref so
  both the editor and the style-measurement HOC share a single DOM node
- **Blocks without** (vote, applause): wrapped edit output in
  `<div {...useBlockProps()}>`
- **cs-embed:** moved `useBlockProps()` from `embed-preview.js` up to
  `edit.js` (called once at the top, before conditional returns); replaced
  `<View>` with `<div {...blockProps}>`
- **poll, cs-embed:** removed `getEditWrapperProps` (redundant with
  `supports.align` + `useBlockProps`)
- **poll/edit.scss:** added `.alignwide` / `.alignfull` class selectors
  alongside the legacy `[data-align]` selectors, since `useBlockProps`
  applies alignment via class names instead of data attributes

Vote block save remains `() => <InnerBlocks.Content />` — this is correct
for dynamic blocks where PHP controls the output.

### Files changed (16 files)

**Block registration** (`apiVersion: 3`):
- `client/blocks/{poll,vote,vote-item,applause,nps,feedback,cs-embed}/index.js`

**Edit components** (`useBlockProps`):
- `client/blocks/{poll,vote,vote-item,applause,nps,feedback,cs-embed}/edit.js`
- `client/blocks/cs-embed/embed-preview.js` (removed useBlockProps, now in parent)

**Styles:**
- `client/blocks/poll/edit.scss`

## Testing

1. Build: `nvm use 18 && pnpm build` — no errors
2. Open the block editor in WordPress 6.9+ or 7.0 — insert each of the 7
   block types and confirm **zero** Block API deprecation warnings in the
   console
3. Edit an existing post containing vote or cs-embed blocks — confirm no
   "block recovery" prompt
4. Set poll alignment to wide and full — confirm padding/border styles apply
   correctly (inspect DOM for `.alignwide`/`.alignfull` classes)

### DIFF
--- a/client/blocks/applause/edit.js
+++ b/client/blocks/applause/edit.js
@@ -55,19 +55,19 @@ const EditApplauseBlock = ( props ) => {
 
 	return (
 		<div { ...blockProps }>
-		<ConnectToCrowdsignal
-			blockIcon={ null }
-			blockName={ __( 'Crowdsignal Applause', 'crowdsignal-forms' ) }
-		>
-			<SideBar
-				{ ...props }
-				shouldPromote={ shouldPromote }
-				signalWarning={ signalWarning }
-				viewResultsUrl={ viewResultsUrl }
-			/>
-			<Toolbar { ...props } />
-			<Applause { ...props } />
-		</ConnectToCrowdsignal>
+			<ConnectToCrowdsignal
+				blockIcon={ null }
+				blockName={ __( 'Crowdsignal Applause', 'crowdsignal-forms' ) }
+			>
+				<SideBar
+					{ ...props }
+					shouldPromote={ shouldPromote }
+					signalWarning={ signalWarning }
+					viewResultsUrl={ viewResultsUrl }
+				/>
+				<Toolbar { ...props } />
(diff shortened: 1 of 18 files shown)


donnchawp (approved):
Works as expected. The warnings about using previous APIs are gone.

PR #307, opened 2026-03-26
Make editor DOM access iframe-safe for Block API v3

stephdau:
## Summary

Depends on #306.

With Block API v3, the editor runs inside an iframe where `window` and
`document` refer to the parent frame, not the editor frame. This PR replaces
all remaining global DOM access in editor code with iframe-safe equivalents
using `ownerDocument` / `defaultView` from element refs.

See [Migrating Blocks for iframe Editor Compatibility](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-api-versions/block-migration-for-iframe-editor-compatibility/).

**feedback/edit.js:**
- Replaced `window.requestAnimationFrame` with `defaultView.requestAnimationFrame`
- Replaced `document.getElementsByClassName` with `ownerDocument.getElementsByClassName`
- Replaced `window.innerHeight` / `window.innerWidth` with `defaultView` equivalents
- Simplified the overlay position guard (removed `ownerDocument !== document`
  check, now uses presence of the content wrapper element)

**components/feedback/util.js:**
- Added `win` parameter to `addFrameOffsets`, `getFeedbackButtonVerticalPosition`,
  and `getFeedbackButtonPosition` so callers can pass the correct window context

**components/poll/util.js:**
- Added `doc` parameter to `loadCustomFont` — replaces `document.head` and
  `document.createElement` with the passed document reference

**blocks/poll/edit.js:**
- Added `blockRef` merged with `blockProps.ref` to provide `ownerDocument`
  to `loadCustomFont`

**components/with-fixed-position/index.js:**
- Guarded `setOffset[clientId]()` call — the registry callback may not be
  ready on the first render cycle in the iframe editor

### Files changed (5 files)

- `client/blocks/feedback/edit.js`
- `client/blocks/poll/edit.js`
- `client/components/feedback/util.js`
- `client/components/poll/util.js`
- `client/components/with-fixed-position/index.js`

## Testing

1. Build: `nvm use 18 && pnpm build` — no errors
2. In WordPress 7.0 (iframe editor): insert a feedback block — confirm it
   positions correctly and no console errors appear
3. Insert a poll block with a custom Google Font — confirm the font loads
   (check `<head>` of the iframe document for the Google Fonts `<link>`)
4. Open and close the feedback popover overlay — confirm overlay dimensions
   match the editor content area

### DIFF
--- a/client/blocks/feedback/edit.js
+++ b/client/blocks/feedback/edit.js
@@ -100,7 +100,8 @@ const EditFeedbackBlock = ( props ) => {
 		// The sidebar itself is rendered off the same setting
 		// hence the need to wait until it's re-rendered
 		// so we can measure it properly.
-		window.requestAnimationFrame( () =>
+		const win = blockElement.current?.ownerDocument.defaultView ?? window;
+		win.requestAnimationFrame( () =>
 			setSidebarChanged( ! sidebarChanged )
 		);
 	}, [ isInserterActive, isListViewActive, isSidebarActive ] );
@@ -109,6 +110,9 @@ const EditFeedbackBlock = ( props ) => {
 		if ( isExample || ! triggerButton.current || widgetEditor ) {
 			return;
 		}
+
+		const doc = blockElement.current.ownerDocument;
+		const win = doc.defaultView;
 
 		setPosition(
 			getFeedbackButtonPosition(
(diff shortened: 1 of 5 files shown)


donnchawp (approved):
It works. The feedback form width can't be changed AFAICT, but I was able to change the font used in a poll and it changed, after my small addition.

